PR #230, opened 2023-02-08
Added clearing of the m.batchInProcess.Store

mtsitrin:
Stopped block production if the block skew too high

### DIFF
--- a/block/manager.go
+++ b/block/manager.go
@@ -35,6 +35,7 @@ type blockSource string
 const (
 	defaultDABlockTime = 30 * time.Second
 	DABatchRetryDelay  = 20 * time.Second
+	maxBlockSkew       = 20000
 )
 
 const (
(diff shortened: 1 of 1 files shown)


block/manager.go
```diff
@@ -575,6 +586,10 @@ func (m *Manager) produceBlock(ctx context.Context) error {
 }
 
 func (m *Manager) submitNextBatch(ctx context.Context) {
+	defer func() {
```
mtsitrin, line 589: @omritoptix this allows the submit process to recover and try again in case of failure

block/manager.go
```diff
@@ -504,6 +505,16 @@ func (m *Manager) produceBlock(ctx context.Context) error {
 	height := m.store.Height()
 	newHeight := height + 1
 
+	syncTarget := atomic.LoadUint64(&m.syncTarget)
+	if newHeight-syncTarget >= maxBlockSkew {
```
mtsitrin, line 509: @omritoptix Not tested. Initial idea to stop the block production.

To **reduce** block production, we can return err here, and modify the timer interval of the caller

omritoptix: Great Idea!
my 2 cents to elaborate on the subject. 
- We need a way to make sure that the skew block size represents the fact that we’ve for sure failed to send a batch and it should be set dynamically. My initial suggestion:
    - define `max success time` as the max time it takes to send a batch successfully which is composed of the following: 
        - sending to the DA (sync)
        - sending the to the HUB
        - Receiving ack it has been accepted by the hub
    - i.e `max success time` = celestia max retry + hub retry time + max time we wait for ack from hub
    - skew size = (`max success time` + epsilon) *  block time
    - we must ensure  `max success time`  < `batch time production` for the following reasons:
        - we will have skew adding up even on successful submissions.
        - we won’t be able to catch up if necessary
    - once we successfully submit the batch again, the skew decreases and the system goes back.

PR #277, opened 2023-03-06
feature: dymint manages the tendermint validator set

mtsitrin:
This PR makes dymint manage the validator set instead of using the App ValidatorUpdates

Dymint ignores the validator updates it will receive on `EndBlock`
It will load the validator set based on the hub on Genesis

### DIFF
--- a/block/manager.go
+++ b/block/manager.go
@@ -93,12 +93,7 @@ func getInitialState(store store.Store, genesis *tmtypes.GenesisDoc) (types.Stat
 		return s, nil
 	}
 
-	s, err = types.NewFromGenesisDoc(genesis)
-	if err != nil {
-		return s, err
-	}
-
-	return s, err
+	return types.NewFromGenesisDoc(genesis)
 }
 
 // NewManager creates new block Manager.
--- a/block/manager.go
+++ b/block/manager.go
@@ -9,6 +9,7 @@ import (
 
 	"code.cloudfoundry.org/go-diodes"
 	"github.com/avast/retry-go"
+	cryptocodec "github.com/cosmos/cosmos-sdk/crypto/codec"
 	abciconv "github.com/dymensionxyz/dymint/conv/abci"
 	"github.com/dymensionxyz/dymint/p2p"
 	"github.com/libp2p/go-libp2p-core/crypto"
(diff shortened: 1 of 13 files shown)


block/manager.go
```diff
@@ -88,10 +89,11 @@ type Manager struct {
...
 func getInitialState(store store.Store, genesis *tmtypes.GenesisDoc) (types.State, error) {
 	s, err := store.LoadState()
-	if err != nil {
-		s, err = types.NewFromGenesisDoc(genesis)
+	if err == nil {
+		return s, nil
 	}
-	return s, err
+
+	return types.NewFromGenesisDoc(genesis)
```
omritoptix, line 96: Seems like if we get an error not related to not having a genesis state, we'll still try to create a genesis state (Like happend to us before) without knowing the root cause of the error until debugging. 
My suggestion is to check for a specific error type if possible before creating `NewFromGensisDoc`

state/executor.go
```diff
@@ -80,7 +74,7 @@ func (e *BlockExecutor) InitChain(genesis *tmtypes.GenesisDoc) (*abci.ResponseIn
 				AppVersion: params.Version.AppVersion,
 			},
 		},
-		Validators:    tmtypes.TM2PB.ValidatorUpdates(tmtypes.NewValidatorSet(validators)),
+		Validators:    tmtypes.TM2PB.ValidatorUpdates(tmtypes.NewValidatorSet(nil)),
```
omritoptix, line 77: why are we through the app passing nil validator to the abci app vs passing the active sequencer as the validator?
i.e the validator_set is solely determined by dymint but still being passed to the app. 

How does the app know how to distribute rewards for relevant sequencers?

omritoptix:
The linter falls. make sure you have the pre-commit hook setup and running to avoid such cases.

state/executor.go
mtsitrin: Currently, the app doesn't handle this
The sequencer on the rollapp is registered by tx / genesis for rewards distribution.

omritoptix:
@mtsitrin is the change backward compatible? did you try upgrading a rollapp with this change?

mtsitrin:
> @mtsitrin is the change backward compatible? did you try upgrading a rollapp with this change?

yes. it will remain with last validatorSet it have

omritoptix (approved):

Issue #241, opened 2023-02-12
Celestia batch tx timeout 


omritoptix:
Though this doesn't creates a logical problem as a retry will occur, it makes the sequencer pay twice and delays the update state to the hub.

Ideally what we should probably want to do is submit it with “sync” mode (i,e not "commit"), make sure there was no submission error, and subscribe to an event vs relaying on "commit" to succeed on first attempt. 

Due to the first timeout the following submission attempts are also failing as we get `tx already in mempool` (code 19) error which cascades to other errors which in turn causes frequent delays in batch submission

![Screen Shot 2023-04-30 at 18 12 31](https://user-images.githubusercontent.com/3949011/235367941-e5bc5ff0-b5d4-4706-b40e-bd840ca0a78a.png)

omritoptix:
Tasks:

- [ ] Update celestia client to support polling 
- [ ] Update go-cnc to a version which always returns txHash (also in the case of error)

PR #301, opened 2023-04-20
fix: changed Backoff delay to Fixed delay

mtsitrin:
# PR Standards

## Opening a pull request should be able to meet the following requirements:

---

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

--- 

After reviewer approval:

### DIFF
--- a/block/manager.go
+++ b/block/manager.go
@@ -37,6 +37,7 @@ const (
 	defaultDABlockTime = 30 * time.Second
 	DABatchRetryDelay  = 20 * time.Second
 	SLBatchRetryDelay  = 10 * time.Second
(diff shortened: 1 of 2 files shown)


omritoptix:
I don't think it's the right direction as we may spam the celestia network though the transaction was already included in the mempool (though not in a block yet) when the attempt between delays is too short.

I believe [this](https://github.com/dymensionxyz/dymint/issues/241) should be the long term solution but until then I think it's better to have either exponential backoffs (just probably with last retries and maybe optimize the `retry_factor`) or alternatively significantly increase the time between fixed retries to make sure we don't send tx submission requests while the previous tx is in the mempool.

PR #321, opened 2023-05-15
feat: added max bytes size check when creating the batch

mtsitrin:
# PR Standards

## Opening a pull request should be able to meet the following requirements:

---

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

--- 

After reviewer approval:


block/manager.go
```diff
@@ -728,9 +735,18 @@ func (m *Manager) createNextDABatch(startHeight uint64, endHeight uint64) (*type
 			m.logger.Error("Failed to load commit", "height", height)
 			return nil, err
 		}
+
+		//Check if the batch size is too big
+		totalSize = totalSize + block.ToProto().Size() + commit.ToProto().Size()
```
omritoptix, line 740: totalSize doesn't equal batch.ToProto().Size() which is what eventually is sent to the DA and the gap gets bigger the bigger the blocks num is.

omritoptix (changes requested):

block/manager.go
mtsitrin: why not?
```
type Batch struct {
	StartHeight uint64
	EndHeight   uint64
	Blocks      []*Block
	Commits     []*Commit
}
```
I want to avoid adding the block to the array and than remove it if batch too large
I think it has the same affect

omritoptix: > why not?

I believe it's the extra encoding or padding added from having a list of blocks/comments in the batch.

Either way it's not equal. you can compare and see. 

For example with 1000 empty blocks the `totalSize`: 519006 and the batch size is 524006

If someone will write maxBatchSize 2MB and it will be more de facto, it will fail with sending the batch.

mtsitrin:
Fixed and added UT

block/manager.go
```diff
@@ -736,14 +735,21 @@ func (m *Manager) createNextDABatch(startHeight uint64, endHeight uint64) (*type
...
+		batch.Blocks = append(batch.Blocks, block)
+		batch.Commits = append(batch.Commits, commit)
+
 		//Check if the batch size is too big
-		totalSize = totalSize + block.ToProto().Size() + commit.ToProto().Size()
-		if totalSize >= int(m.conf.BlockBatchSizeBytes) {
+		totalSize := batch.ToProto().Size()
+		if m.conf.BlockBatchSizeBytes > 0 && totalSize >= int(m.conf.BlockBatchSizeBytes) {
+			// Nil out the last block and commit
+			batch.Blocks[len(batch.Blocks)-1] = nil
```
omritoptix, line 745: Is setting it to nil it before removing it is just for the garbage collector or is there any other reason?

block/manager_test.go
```diff
@@ -404,12 +403,78 @@ func TestBatchRetryWhileBatchAccepted(t *testing.T) {
...
 	blockLoopContext, blockLoopCancel = context.WithCancel(context.Background())
 	defer blockLoopCancel()
 	go manager.ProduceBlockLoop(blockLoopContext)
 	time.Sleep(1 * time.Second)
-	assert.Greater(atomic.LoadUint64(&settlementLayerClient.batchCounter), uint64(0))
+	assert.Greater(atomic.LoadUint64(&settlementLayerClient.BatchCounter), uint64(0))
+
+}
+
+func Test_createNextDABatchWithBytesLimit(t *testing.T) {
```
omritoptix, line 415: So far our testing convention in dymint is pascal case. 
not sure if the underscore is by mistake or a different convention by I rather we keep it unified and use pascal case.

block/manager_test.go
```diff
@@ -404,12 +403,78 @@ func TestBatchRetryWhileBatchAccepted(t *testing.T) {
...
+	require.NoError(err)
+	// Init manager
+	managerConfig := getManagerConfig()
+	managerConfig.BlockBatchSize = 1000
+	managerConfig.BlockBatchSizeBytes = 2000 //enough for 2 block, not enough for 10 blocks
+	manager, err := getManager(managerConfig, nil, nil, 1, 1, 0, proxyApp, nil)
+	require.NoError(err)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer func() {
```
omritoptix, line 432: why not simply `defer cancel()`?

block/manager_test.go
```diff
@@ -404,12 +403,78 @@ func TestBatchRetryWhileBatchAccepted(t *testing.T) {
...
+	assert.Less(batch.ToProto().Size(), int(managerConfig.BlockBatchSizeBytes))
+
+	//Now produce 10 blocks and batch them assuming they don't fit in the bytes limit
+	blocksToProduce = 10
+	for i := 0; i < blocksToProduce; i++ {
+		manager.produceBlock(ctx)
+	}
+
+	// Call createNextDABatch function
+	startHeight = atomic.LoadUint64(&manager.syncTarget) + 1
```
omritoptix, line 458: possibly turning it to table tests will prevent some code repitition.

block/manager.go
mtsitrin: yes, probably not mandatory, but to be on the safe side (as the array is of pointers)
otherwise, the "deleted" elements will still exist as long as the "batch" struct exists

block/manager.go
```diff
@@ -736,14 +735,21 @@ func (m *Manager) createNextDABatch(startHeight uint64, endHeight uint64) (*type
...
 		}
 
+		batch.Blocks = append(batch.Blocks, block)
+		batch.Commits = append(batch.Commits, commit)
+
 		//Check if the batch size is too big
-		totalSize = totalSize + block.ToProto().Size() + commit.ToProto().Size()
-		if totalSize >= int(m.conf.BlockBatchSizeBytes) {
+		totalSize := batch.ToProto().Size()
+		if m.conf.BlockBatchSizeBytes > 0 && totalSize >= int(m.conf.BlockBatchSizeBytes) {
```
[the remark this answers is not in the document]
mtsitrin: Actually, It's not the same.
If the flag not provided, it gets the default value by the `Viper` conf

If it's set explicitly to `0` - I thought it should mean `no limit`. U say it should enforce the default as in line 121.

Fine, changing it.

block/manager.go
```diff
@@ -117,10 +116,17 @@ func NewManager(
...
 	// TODO(omritoptix): Think about the default batchSize and default DABlockTime proper location.
 	if conf.DABlockTime == 0 {
-		logger.Info("WARNING: using default DA block time", "DABlockTime", defaultDABlockTime)
-		conf.DABlockTime = defaultDABlockTime
+		logger.Info("WARNING: using default DA block time", "DABlockTime", config.DefaultNodeConfig.DABlockTime)
+		conf.DABlockTime = config.DefaultNodeConfig.DABlockTime
+	}
+
+	if conf.BlockBatchSizeBytes == 0 {
+		logger.Info("WARNING: using default DA batch size bytes limit", "DABlockTime", config.DefaultNodeConfig.BlockBatchSizeBytes)
```
omritoptix, line 128: typo. it says "DABlockTime" in the log message (I guess meant to be "BatchSize").

omritoptix (approved):

PR #327, opened 2023-05-21
feat: propagate node health to RPC server response

omritoptix:
# PR Standards

---

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/rpc/server.go
+++ b/rpc/server.go
@@ -11,30 +11,48 @@ import (
 	"github.com/rs/cors"
 	"github.com/tendermint/tendermint/config"
 	"github.com/tendermint/tendermint/libs/log"
+	"github.com/tendermint/tendermint/libs/pubsub"
 	"github.com/tendermint/tendermint/libs/service"
 	rpcclient "github.com/tendermint/tendermint/rpc/client"
 	"golang.org/x/net/netutil"
 
 	"github.com/dymensionxyz/dymint/node"
+	"github.com/dymensionxyz/dymint/node/events"
 	"github.com/dymensionxyz/dymint/rpc/client"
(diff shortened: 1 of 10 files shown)


rpc/server.go
```diff
@@ -11,30 +11,48 @@ import (
...
+	healthStatus sharedtypes.HealthStatus
+	listener     net.Listener
+	ctx          context.Context
 
 	server http.Server
 }
 
 // NewServer creates new instance of Server with given configuration.
-func NewServer(node *node.Node, config *config.RPCConfig, logger log.Logger) *Server {
+func NewServer(node *node.Node, config *config.RPCConfig, logger log.Logger, listener ...net.Listener) *Server {
```
mtsitrin, line 43: as the `listener` arg used only for testing, I would have rather move it to separate `WithListener` method

rpc/server.go
```diff
@@ -11,30 +11,48 @@ import (
...
 	srv := &Server{
 		config: config,
 		client: client.NewClient(node),
+		node:   node,
+		healthStatus: sharedtypes.HealthStatus{
```
mtsitrin, line 48: why not implementing it as part of the middleware? to avoid changing the base server implementaion

omritoptix: as I see it the middleware should be very flat, light and reusable and only get the health status and react according to it.

I see the server is the orchestrator, hence it listens to events and changes relevant variables.

If I'd implement it as part of the middleware it would be too opinionated, non reusable and will need to be passed with an event bus and subscribe to events etc.

mtsitrin (approved):

PR #328, opened 2023-05-22
feat: add pruning mechanism that deletes old blocks and commits

mtsitrin:
# PR Standards

---

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


store/pruning.go
```diff
@@ -0,0 +1,75 @@
...
+
+import "fmt"
+
+// PruneBlocks removes block up to (but not including) a height. It returns number of blocks pruned.
+func (s *DefaultStore) PruneBlocks(heightInt int64) (uint64, error) {
+	if heightInt <= 0 {
+		return 0, fmt.Errorf("height must be greater than 0")
+	}
+
+	//TODO: validate the height is > committed/synced height
```
omritoptix, line 11: what does this TODO mean? not sure I understand

store/pruning.go
```diff
@@ -0,0 +1,75 @@
...
+
+		pruned++
+
+		// flush every 1000 blocks to avoid batches becoming too large
+		if pruned%1000 == 0 && pruned > 0 {
+			err := flush(batch, h)
+			if err != nil {
+				return 0, err
+			}
+			//FIXME: I think it mess with the defer above as it overwrites the batch variable
```
omritoptix, line 64: don't think there is a problem here. it will simply defer both functions. it doesn't override the previous defer.

state/executor_test.go
```diff
@@ -210,7 +210,11 @@ func TestApplyBlock(t *testing.T) {
 
 	// Apply the block with an invalid commit
 	err = executor.Validate(state, block, invalidCommit, proposer)
-	require.Error(types.ErrInvalidSignature)
+
+	// FIXME: This test didn't check for specific error. It was just checking for error.
```
omritoptix, line 214: not sure I understand. you're saying that `require.Error(types.ErrInvalidSignature)` failed? if so how did the test pass? or what am I missing?

node/node.go
```diff
@@ -138,6 +138,7 @@ func NewNode(ctx context.Context, conf config.NodeConfig, p2pKey crypto.PrivKey,
 		baseKV = store.NewDefaultKVStore(conf.RootDir, conf.DBPath, "dymint")
 	}
 	mainKV := store.NewPrefixKV(baseKV, mainPrefix)
+	// FIXME: not sure why dalcKV is needed. It's not used by celestia
```
omritoptix, line 141: it's currently used by the mock but perhaps there is a more elegant way to transfer it. either way with current design it's needed.

block/pruning.go
```diff
@@ -0,0 +1,17 @@
+package block
+
+import "fmt"
+
+func (m *Manager) pruneBlocks(retainHeight int64) (uint64, error) {
+	if retainHeight <= int64(m.store.Base()) {
```
omritoptix, line 6: I think this check is also done in the m.store.PruneBlocks method

store/pruning.go
```diff
@@ -0,0 +1,75 @@
...
+		return 0, fmt.Errorf("cannot prune to height %v, it is lower than base height %v",
+			height, base)
+	}
+
+	pruned := uint64(0)
+	batch := s.db.NewBatch()
+	defer batch.Discard()
+
+	flush := func(batch Batch, base uint64) error {
+		s.SetBase(base)
```
omritoptix, line 27: shouldn't the `setBase` get called only if the batch.Commit() succeeds?

store/store.go
```diff
@@ -58,6 +59,19 @@ func (s *DefaultStore) Height() uint64 {
...
+// SetBase sets the height saved in the Store of the earliest block
+func (s *DefaultStore) SetBase(height uint64) {
+	baseHeight := atomic.LoadUint64(&s.baseHeight)
+	if height > baseHeight {
+		_ = atomic.CompareAndSwapUint64(&s.baseHeight, baseHeight, height)
+	}
+}
+
+// Base returns height of the earliest block saved in the Store.
+func (s *DefaultStore) Base() uint64 {
```
omritoptix, line 71: Not sure I see where base is initialized after a restart? Maybe I missed that. 
So for example if base is 100, and I restart the process, it seems as if it will return to be 0.
For example the height is saved in the state and initialized in store.go line 210 under LoadState() function.

state/executor_test.go
mtsitrin: `require.Error(types.ErrInvalidSignature)` is wrong. it checks that the type `types.ErrInvalidSignature` is of `error` type.

checking for specific error using `require.ErrorIs` fails.

I've changed it to check for any error
`require.Error(err)`

if u check
require.ErrorIs

store/store.go
mtsitrin: u right. I've tried to avoid changes to the state, and the `0-100` blocks in your example will be skipped when pruning for the first time after restert.
```
for h := base; h < height; h++ {
		hash, err := s.loadHashFromIndex(h)
		if err != nil {
			continue
		}
```

U think it should be added to the state?

omritoptix: Actually the major problem I'm seeing here is that if base=0 after restart, and `retainHeight` returned from the app in the [pruning.go](https://github.com/dymensionxyz/dymint/pull/328/commits/2509b69582d1b51edd7b7a2f8f43cda140df5735#diff-812ea47bd8817fdc1e28a12177c62dfb338c3679ae9e0a132532ac4e09d83db8) is bigger, then in line 6 seems like it will never actually call `store.pruneBlocks` and as a result will never update the base

omritoptix:
there are 2 comments I've left which seems like are still open

store/store.go
omritoptix: you're right. my miss here. 

but still it seems like if the `retainHeight` is in the order of millions and we restart the rollapp and base is 0, so  this will be iterated for a million times without a reason:

```
for h := base; h < height; h++ {
		hash, err := s.loadHashFromIndex(h)
		if err != nil {
			continue
		}
```

mtsitrin: U correct, but I'm not sure it's worth the effort now to keep the Base in the persistent store.

I would leave it as enhancement, + adding pruning in intervals which currently not implemented
(i.e a pruning will happen after each Commit)

If u think it should be handled in this PR, let me know
Otherwise, approve and I'll open a separate issue

omritoptix: I think it's relatively a small effort to add the base to the persisted state and IMO it's very likely that rollups will restart after having a huge height which will in turn cause millions of redundant I/O calls. 

IMO the feature is not complete as I don't see persisting the base as an enhancement but as a bug if it doesn't exist (as base is expected to be something else upon restart). 

Re pruning intervals, I agree it's an enhancement.

omritoptix (approved):

PR #331, opened 2023-05-30
fix: removed panic alert from graceful shutdown

mtsitrin:
# PR Standards

---

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/settlement/dymension/dymension.go
+++ b/settlement/dymension/dymension.go
@@ -249,6 +249,7 @@ func (d *HubClient) eventHandler() {
 	for {
 		select {
 		case <-d.ctx.Done():
+			return
 		case <-d.client.EventListenerQuit():
 			return
 		case event := <-eventsChannel:
--- a/settlement/dymension/dymension.go
+++ b/settlement/dymension/dymension.go
@@ -158,11 +158,12 @@ func (d *HubClient) Start() error {
 
 // Stop stops the HubClient.
 func (d *HubClient) Stop() error {
+	d.cancel()
 	err := d.client.StopEventListener()
 	if err != nil {
 		return err
 	}
-	d.cancel()
+
 	return nil
 }
 
@@ -251,7 +252,7 @@ func (d *HubClient) eventHandler() {
 		case <-d.ctx.Done():
 			return
 		case <-d.client.EventListenerQuit():
-			return
+			panic("Settlement WS disconnected")
 		case event := <-eventsChannel:
 			// Assert value is in map and publish it to the event bus
 			d.logger.Debug("Received event from settlement layer")
--- a/settlement/dymension/dymension.go
+++ b/settlement/dymension/dymension.go
@@ -249,9 +249,8 @@ func (d *HubClient) eventHandler() {
 	for {
 		select {
 		case <-d.ctx.Done():
-			return
 		case <-d.client.EventListenerQuit():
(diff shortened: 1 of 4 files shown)


settlement/dymension/dymension.go
```diff
@@ -249,9 +249,8 @@ func (d *HubClient) eventHandler() {
 	for {
 		select {
 		case <-d.ctx.Done():
-			return
 		case <-d.client.EventListenerQuit():
-			panic("Settlement WS disconnected")
+			return
```
omritoptix, line 254: not sure I understand the logic of `return` here. 
 
as I see it if we `return` here vs panicing, we don't have a websocket connection to the hub anymore, hence we won't get events of update hence we'll never move to submitting the next batch. so basically the rollapp will continue creating batches without submitting them.

Panic in that case makes the rollapp restart and retry the connection to the hub.

Either way don't think panic is the long term solution here, but I'm actually working on a reconnect solution.

omritoptix (commented):
Maybe the issue wasn't clear in terms of the requirements, but at least what I meant is catching panic in critical locations and gracefully shutting down the goroutines by cancelling their context. so for example inside the main node `start` function we can do:

```
defer func() {
        if r := recover(); r != nil {
            fmt.Println("Recovered from panic", r)
            // Cancel the context, triggering shutdown of worker goroutines
            cancel()
        }
    }()
    
    // Possibly use a waitGroup start all the goroutines and wait for all the them to be done. before exiting
    wg.Wait()
```

or something alone those lines.

settlement/dymension/dymension.go
mtsitrin: The issue as I see it, is that for graceful shutdown using SIGTERM, no panic should happen. 
But U totally right. It should panic when websocket with the hub is disconnected.
I Will fix the PR to support both requirements.

Regarding the panic through the code,
Panic can be recovered in the same `goroutine` so it's hard to have a single `recover` func that covers all the panics.
IMO, panic in the code should be **only** in unrecoverable logic flow.
We can have a separate issue to review all the panics currently in-place

omritoptix: true. we should handle graceful shutdown using SIGTERM. With that being said I believe we should also handle panic with the same urgency is specific places.  

> Panic can be recovered in the same goroutine so it's hard to have a single recover func that covers all the panics.

Not sure if that's what you meant but panic bubbles up so basically can be caught in the upper most function (i.e node.go). from there the node can wait for all of it's children goroutines to gracefully shutdown - or at least the critical ones.

The reason I think it's important is because we had a few situations where the rollapp kept panicing and some of the panics lead to an inconsistent unrecoverable state in the rollapp which required manual intervention. 

I believe the critical goroutines to make sure are gracefully shutdown are in the `manager`.  so IMO we can start by at least making sure this gracefully shuts down upon panic/SIGTERM.

omritoptix (commented):
Great. 

one thing is missing for me though which IMO is the most important thing which is stopping the manager upon node.stop(). 

Currently the manager doesn't have a stop method so I think we should add it and call when when stopping the node as the manager go routines were usually the ones causing inconsistent state.

mtsitrin:
The manager uses `Node`'s context
All the manager's go routines selecting on this context

omritoptix (approved):

PR #332, opened 2023-05-30
feat: toggle settlement health event upon settlement batch submission status

omritoptix:
## Overview
This pr moves the batch submission to settlement layer logic from the manager to the settlement.
By that it allows us to encapsulate the logic and handle any possible failures.
The manager will only be notified upon submission success or health status changes. 
 
# PR Standards

---

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/block/manager.go
+++ b/block/manager.go
@@ -75,10 +75,7 @@ type Manager struct {
 
 	syncTargetDiode diodes.Diode
 
-	batchInProcess   atomic.Value
-	batchRetryCtx    context.Context
-	batchRetryCancel context.CancelFunc
-	batchRetryMu     sync.RWMutex
+	batchInProcess atomic.Value
 
 	shouldProduceBlocksCh chan bool
 
@@ -366,7 +363,7 @@ func (m *Manager) SyncTargetLoop(ctx context.Context) {
 			// TODO(omritoptix): Once we have leader election, we can add a condition.
 			// Update batch accepted is only relevant for the aggregator
 			// TODO(omritoptix): Check if we are the aggregator
-			m.updateBatchAccepted()
+			m.batchInProcess.Store(false)
 		case <-subscription.Cancelled():
 			m.logger.Info("Subscription canceled")
 		}
@@ -378,15 +375,6 @@ func (m *Manager) updateSyncParams(ctx context.Context, endHeight uint64) {
 	m.logger.Info("Received new syncTarget", "syncTarget", endHeight)
 	atomic.StoreUint64(&m.syncTarget, endHeight)
 	m.syncTargetDiode.Set(diodes.GenericDataType(&endHeight))
-}
-
-func (m *Manager) updateBatchAccepted() {
-	m.batchRetryMu.Lock()
-	if m.batchRetryCtx != nil && m.batchRetryCtx.Err() == nil {
-		m.batchRetryCancel()
-	}
-	m.batchRetryMu.Unlock()
-	m.batchInProcess.Store(false)
 }
 
 // RetriveLoop listens for new sync messages written to a ring buffer and in turn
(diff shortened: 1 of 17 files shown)


block/manager.go
```diff
@@ -720,7 +708,7 @@ func (m *Manager) submitNextBatch(ctx context.Context) {
 	// Submit batch to SL
 	// TODO(omritoptix): Handle a case where the SL submission fails due to syncTarget out of sync with the latestHeight in the SL.
 	// In that case we'll want to update the syncTarget before returning.
-	m.submitBatchToSL(nextBatch, resultSubmitToDA)
+	go m.settlementClient.SubmitBatch(nextBatch, m.dalc.GetClientType(), resultSubmitToDA)
```
mtsitrin, line 711: U have a different style with the DA layer and the SL layer
The DA layer has `m.submitBatchToDA` which manages the retries and the result
While now the SL layer is managed entirely from `m.settlementClient`

settlement/base.go
```diff
@@ -93,27 +93,11 @@ func (b *BaseLayerClient) Stop() error {
 	return nil
 }
 
-// SubmitBatch submits the batch to the settlement layer. This should create a transaction which (potentially)
-// triggers a state transition in the settlement layer.
-func (b *BaseLayerClient) SubmitBatch(batch *types.Batch, daClient da.Client, daResult *da.ResultSubmitBatch) *ResultSubmitBatch {
+// SubmitBatch tries submitting the batch in an async way to the settlement layer. Events are emitted on success or failure.
```
mtsitrin, line 96: How is that `async`?
You still call `d.client.BroadcastTx` and waits to see the tx result

settlement/base.go
```diff
@@ -159,14 +143,13 @@ func (b *BaseLayerClient) fetchSequencersList() ([]*types.Sequencer, error) {
 	return sequencers, nil
 }
 
-func (b *BaseLayerClient) validateBatch(batch *types.Batch) error {
+func (b *BaseLayerClient) validateBatch(batch *types.Batch) {
```
mtsitrin, line 146: Not sure why to remove the `error`
I don't think it's good idea to have panic scattered around

The internal functions should return errors, to be aggregated and panic upon if needed

settlement/dymension/dymension.go
```diff
@@ -262,26 +345,28 @@ func (d *HubClient) eventHandler() {
...
-		return nil, err
+		d.logger.Error("Error getting account for building transaction", "account", d.config.DymAccountName)
+		panic(err)
 	}
 
 	addr, err := account.Address(addressPrefix)
 	if err != nil {
-		return nil, err
+		d.logger.Error("Error getting address prefix for building transaction", "account", addressPrefix)
+		panic(err)
```
mtsitrin, line 369: same here
I guess this error could be checked by the caller and panic

mtsitrin (approved):
I don't like the excessive usage of `panic` instead of returning `errors`,
IMO it's bad practice

block/manager.go
omritoptix: true. that's the [next task ](https://github.com/dymensionxyz/dymint/issues/314) to move the DA submission logic there.

omritoptix:
> I don't like the excessive usage of `panic` instead of returning `errors`, IMO it's bad practice

I see. don't think it's about liking or not nor that it's a bad practice. I it's about the use case for panic. 
For me panicing is in a case where condition is met that cannot be or should not be handled, such as a bug in the code which the process can't recover from.

But let me know which places specifically you think I should have used errors instead of panic so can I check.

settlement/base.go
omritoptix: it's async by the [cosmos terms](https://tutorials.cosmos.network/academy/2-cosmos-concepts/3-transactions.html). submitting a tx in a broadcast mode async in cosmos means submitting the tx and not waiting to see if it was included in a block (but do get a response back). I can make it more clear that it's broadcast mode = async if you think it makes more sense.

settlement/base.go
omritoptix: makes sense

ItayLevyOfficial (approved):

Issue #324, opened 2023-05-17
block batch submission should be determined by max batch size or max hub blocks


omritoptix:
The batch size should only be set by the number of bytes regardless to the number of blocks.
it causes confusion by having 2 different params controlling essentially the same result.

mtsitrin:
MAX BYTES is used as the MAX limit

Batch submission should have some minimum enforcement as well
(otherwise, u can't predict when the batch will reach its max size and will be sent)

Currently, the number of blocks per batch is used as "minimum" enforcement
When empty blocks will be handled, and some minimal block time interval will take place, we could remove the number of blocks 

@omritoptix WDYT?

omritoptix:
update the title to reflect the latest discussion in the subject.
Batch submission should be either when reaching max bytes or alternatively the time (possibly measured by hub blocks) since last batch submission > threshold.

mtsitrin:
Using the max bytes as a trigger to submit is major effort
as it requires keeping the accumulated size and refactors throughout the code

I suggest leaving it as now:
* assuming mostly empty blocks -> the submission interval will be capped by the `submissionTimeout`
* assuming many txs and large blocks -> submission interval will be `batch size` * `block time`

so we have a guarantee of
`batch size` * `block time` < submit interval < `submissionTimeout`


I'm not sure an enhancement beyond it is worth the `High` priority now

mtsitrin:
related to #418

omritoptix:
this becomes a priority since we can relay on interval for the following reasons:

### KPIS
1. keep SL synced (max 1 batch behind) for syncing from DA purposes 
2. keep batch size ideal for DA and for pricing purposes
3. submission rate (byte/second) > creation reate (byte/second)

### Problem with batch interval

The batch time interval, since not depends on block batch size has the following problems:
1. high load +  high interval (e.g 1m40sec like default), given batch size limitation - a gap is formed with submitting to SL (i.e kpi 1 is violated)
2. low load + high interval (e.g 1m40sec) - most batches to da may contain 1 block which viloates kpi 2 
3. high load + low internal (e.g 2 sec) - will violate kip 3

In general, as the system load is ever changing, working with interval is not sustaible. hence we need to change to block size which in essence should automatically set the interval by the load.

PR #342, opened 2023-06-10
feat: Support skipping empty blocks production

mtsitrin:
This PR introduces the feature of `skipping empty blocks production`
The feature is `off` by default

# PR Standards

---

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/block/manager.go
+++ b/block/manager.go
@@ -317,19 +317,26 @@ func (m *Manager) ProduceBlockLoop(ctx context.Context) {
 	}
 
 	ticker := time.NewTicker(m.conf.BlockTime)
-	tickerEmptyBlocksMaxTime := time.NewTicker(m.conf.EmptyBlocksMaxTime)
 
-	//FIXME: we probably want this feature to be toggleable
+	var tickerEmptyBlocksMaxTime *time.Ticker
+	var tickerEmptyBlocksMaxTimeCh <-chan time.Time
+	if m.conf.EmptyBlocksMaxTime > 0 {
+		tickerEmptyBlocksMaxTime = time.NewTicker(m.conf.EmptyBlocksMaxTime)
+		tickerEmptyBlocksMaxTimeCh = tickerEmptyBlocksMaxTime.C
+	}
 
 	//Allow the initial block to be empty
 	produceEmptyBlock := true
 	for {
 		select {
+		//Context canceled
 		case <-ctx.Done():
 			return
-		case <-tickerEmptyBlocksMaxTime.C:
-			m.logger.Error("No transactions for too long, producing empty block")
+		//Empty blocks timeout
+		case <-tickerEmptyBlocksMaxTimeCh:
+			m.logger.Error("No transactions for too long, allowing to produce empty block")
 			produceEmptyBlock = true
+		//Produce block
 		case <-ticker.C:
 			err := m.produceBlock(ctx, produceEmptyBlock)
 			if err == types.ErrSkippedEmptyBlock {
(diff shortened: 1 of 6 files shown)


block/production_test.go
```diff
@@ -0,0 +1,158 @@
...
+	//Check initial height
+	expectedHeight := uint64(0)
+	assert.Equal(expectedHeight, manager.store.Height())
+
+	mCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	go manager.ProduceBlockLoop(mCtx)
+
+	<-time.Tick(1 * time.Second)
+	err = mpool.CheckTx([]byte{1, 2, 3, 4}, nil, mempool.TxInfo{})
```
mtsitrin, line 131: @omritoptix If you can take a look
I can't manage to send a tx

all the blocks produced are empty
The test is skipped for now. I've tested it manually

block/manager.go
```diff
@@ -308,23 +308,49 @@ func (m *Manager) ProduceBlockLoop(ctx context.Context) {
...
+	//Allow the initial block to be empty
+	produceEmptyBlock := true
 	for {
 		select {
+		//Context canceled
 		case <-ctx.Done():
 			return
+		//Empty blocks timeout
+		case <-tickerEmptyBlocksMaxTimeCh:
+			m.logger.Error("No transactions for too long, allowing to produce empty block")
```
omritoptix, line 327: not sure I understand why would I want to produce an empty block if there were not transactions for too long (assuming the hub can receive empty batches for liveness)?

block/production_test.go
omritoptix: Will do.

config/defaults.go
```diff
@@ -19,8 +19,9 @@ var DefaultNodeConfig = NodeConfig{
 	Aggregator: true,
 	BlockManagerConfig: BlockManagerConfig{
 		BlockTime:           200 * time.Millisecond,
+		EmptyBlocksMaxTime:  0,
```
omritoptix, line 22: IMO should be enabled by default. Actually don't think it should even be a flag. thinking it should simply not produce empty blocks hardcoded but maybe I'm missing something. 

is there any reason you think an operator would like to produce empty blocks assuming no cron job depends on it?

block/manager.go
mtsitrin: IMO It makes more sense to separate `block production` and `batch submission`
Block production is important for rollapp liveness
while the batch submission is more related to finality

I will add an `expiry timeout` for the batch submission trigger as well

So in a sense,
block_time < EmptyBlocksMaxTime << BatchSubmissionMaxTime
(e.g 200ms < 1m << 1h)

config/defaults.go
mtsitrin: In case some modules have begin block / end block logic that needs to run even on empty blocks
(e.g epoch, which checks the timers and call the hooks on begin block)

block/manager.go
mtsitrin: It will be handled in this issue
https://github.com/dymensionxyz/dymint/issues/324

omritoptix: good thinking with the `batchSubmissionMaxTime`. 

though I think it's already necessary.  if it doesn't exist now (and only batch size does), so it will take the sequencer (assuming empty blocks produced every 1 minute) hours to update the hub and may hurt the uptime as indicated by the hub/portal.  

btw we may do `batchSubmissionMaxBlocks` which measures time by hub blocks.  

I think it makes more sense as this is how the rollapp update rate for slashing etc is going to be checked by the hub.

PR #347, opened 2023-06-13
feat: block batch submission limited with timeout

mtsitrin:
# PR Standards

---

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


block/manager.go
```diff
@@ -707,8 +711,9 @@ func (m *Manager) produceBlock(ctx context.Context, allowEmpty bool) error {
 
 func (m *Manager) submitNextBatch(ctx context.Context) {
 	// Get the batch start and end height
-	startHeight := atomic.LoadUint64(&m.syncTarget) + 1
-	endHeight := startHeight + m.conf.BlockBatchSize - 1
+	startHeight := atomic.LoadUint64(&m.syncTarget)
+	endHeight := uint64(m.lastState.LastBlockHeight)
```
omritoptix, line 736: If we are in a case where batch submission takes a long time, or the base layer RPC is down, the aggregator may produce more blocks then the batch size (even though we are supposed to stop producing blocks on rpc down, but we can't count on how long it will take). 

In such situation the batch size, i.e `lastState.LastBlockHeight - startHeight`, can be more than max size of blocks which can cause problems sending base layers. 

a possible solution would be to set the `endHeight := min(startHeight + batchSize + 1, m.lastState.LastBlockHeight)`.

block/manager.go
```diff
@@ -276,6 +270,7 @@ func (m *Manager) waitForSync(ctx context.Context) error {
 		// the SL still hasn't got any batches for this chain.
 		m.logger.Info("No batches for chain found in SL. Start writing first batch")
 		atomic.StoreUint64(&m.syncTarget, uint64(m.genesis.InitialHeight-1))
+		atomic.StoreInt64(&m.lastSubmissionTime, time.Now().Unix())
```
omritoptix, line 281: Seems to me like only if we start a fresh sequencer the `lastSubmissionTime` will get initialized, otherwise if we restart a sequencer which already produced blocks, the `lastSubmissionTime` will not be initialized. 

Why not just put it in the beginning of the `produceBlockLoop`?

block/manager.go
mtsitrin: the batch size is "fictive", it's just triggered to start batch submission
when creating the batch itself, we'll try to fill it until it reaches lastHeight or size limit

even if `endHeight - startHeight == 2000000`, it will iterate over the blocks until it reaches max size.

block/manager.go
mtsitrin: It's also in the `updateSyncParams` so it should cover all non-fresh executions
I'll move it to the `produceBlockLoop` anyway

block/manager.go
omritoptix: Not sure I understand what you're saying or I'm missing something but In the case I described the `endHeight` is not fictive, but real. 

it's already after the batch submission has been triggered.  

thing is that `m.lastState.LastBlockHeight` you described as lastHeight can be much bigger than the desired `endHeight` if we continue producing block while batch submission was stuck. 

what am I missing?

block/manager.go
omritoptix: true. but if we get an error (different then `ErrBatchNotFound`) in 'waitForSync' the  `updateSyncParams` will not be called but we'll still continue producing blocks (despite the error). 

don't remember the reason for that (maybe a bug) but this is the current behavior.

omritoptix (approved):

PR #367, opened 2023-06-20
feat: mock SL doesn't use keyring path from config. 

mtsitrin:
mock SL doesn't use keyring path from config
this allows us to set whenever keyring_dir for the sequencer as default

set defaults to be more clear and useful for local run

The default path for sequencer keys will be `rootDir/sequecner_keys`

# PR Standards

---

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -3,7 +3,6 @@ package node
 import (
 	"context"
 	"encoding/base64"
-	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"sync"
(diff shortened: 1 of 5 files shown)


node/node.go
```diff
@@ -157,6 +158,13 @@ func NewNode(ctx context.Context, conf config.NodeConfig, p2pKey crypto.PrivKey,
 	if settlementlc == nil {
 		return nil, fmt.Errorf("couldn't get settlement client named '%s'", conf.SettlementLayer)
 	}
+	if conf.SettlementLayer == "mock" {
```
omritoptix, line 161: do we have a way to make sure that making this check against mock is not done in our non mock related code? seems a bit dirty to me.

node/node.go
```diff
@@ -159,11 +158,7 @@ func NewNode(ctx context.Context, conf config.NodeConfig, p2pKey crypto.PrivKey,
 		return nil, fmt.Errorf("couldn't get settlement client named '%s'", conf.SettlementLayer)
 	}
 	if conf.SettlementLayer == "mock" {
-		pubKeybytes, err := signingKey.GetPublic().Raw()
-		if err != nil {
-			return nil, err
-		}
-		conf.SettlementConfig.ProposerPubKey = hex.EncodeToString(pubKeybytes)
+		conf.SettlementConfig.KeyringHomeDir = conf.RootDir
```
omritoptix, line 161: why do we need to initialize the `KeyringHomeDir` here  vs just taking it from the dymint.toml? we have `keyring_home_dir` in the dymint.toml.

mtsitrin:
Because I want the default to be other than root dir
And mock SL expects it to be the root

PR #373, opened 2023-06-26
fix: fixed namespaceID parsing from toml

mtsitrin:
# PR Standards

---

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/da/celestia/config.go
+++ b/da/celestia/config.go
@@ -1,0 +1,68 @@
+package celestia
+
+import (
+	"encoding/hex"
+	"encoding/json"
+	"time"
+)
+
+const (
+	defaultTxPollingRetryDelay = 20 * time.Second
+	defaultSubmitRetryDelay    = 10 * time.Second
+	defaultTxPollingAttempts   = 5
+)
+
+// Config stores Celestia DALC configuration parameters.
+type Config struct {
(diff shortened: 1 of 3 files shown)


da/celestia/config.go
```diff
@@ -0,0 +1,68 @@
...
+	BaseURL     string        `json:"base_url"`
+	AppNodeURL  string        `json:"app_node_url"`
+	Timeout     time.Duration `json:"timeout"`
+	Fee         int64         `json:"fee"`
+	GasLimit    uint64        `json:"gas_limit"`
+	NamespaceID [8]byte       //`json:"namespace_id"`
+}
+
+// Define an auxiliary type to prevent infinite recursion in UnmarshalJSON
+type auxConfig struct {
```
omritoptix, line 26: don't really like the duplication here. if I understand correctly the goal is to parse namespace id from string to bytes. I believe previously it has been done in the config file. 
why do you think using AuxConfig is preferred? 
Another option is even to a a field to the main config. Creating 2 config which are basically the same but one fields seems very confusing to me.

mtsitrin: Yeah I'll figure out better way

omritoptix (approved):

PR #408, opened 2023-07-18
feat: Introduce Basic HTTP Metrics Server with rollapp height and hub rollapp Height 

ItayLevyOfficial:
# PR Standards

---

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"sync"
 
 	"github.com/libp2p/go-libp2p/core/crypto"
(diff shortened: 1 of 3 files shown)


omritoptix:
@ItayLevyOfficial  can pull new changes with main. should fix the failed tests.

node/node.go
```diff
@@ -400,5 +405,22 @@ func (n *Node) healthStatusHandler(err error) {
...
+		http.Handle("/metrics", promhttp.Handler())
+		srv := &http.Server{
+			Addr:         ":2112",
+			ReadTimeout:  5 * time.Second,
+			WriteTimeout: 10 * time.Second,
+			Handler:      http.DefaultServeMux,
+		}
+		if err := srv.ListenAndServe(); err != nil {
+			// Error starting or closing listener:
+			n.Logger.Error("Prometheus HTTP server ListenAndServe", "err", err)
```
omritoptix, line 423: here seems like we swallow the error - i.e the sequencer will start even if the metric server will not which IMO may be confusing for the user.
Why not catch the error, return to the caller, and have the caller panic on it?

omritoptix:
Also IMO that it would be cleaner to have 1 issue for adding a metrics server and another issue for adding specific metrics vs doing it in the same issue/PR.

PR #417, opened 2023-07-22
feat: celestia fee should be calculated dynamically

mtsitrin:
# PR Standards

---

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/da/celestia/celestia.go
+++ b/da/celestia/celestia.go
@@ -174,7 +174,6 @@ func (c *DataAvailabilityLayerClient) SubmitBatch(batch *types.Batch) da.ResultS
 			c.logger.Debug("Context cancelled")
 			return da.ResultSubmitBatch{}
 		default:
-			//The gas needed for the batch probably should be computed before, when creating the batch. similiar to max bytes check
 			estimatedGas := float64(EstimateGas([]int{len(blob)}))
 			gasWanted := uint64(estimatedGas * gasAdjustment)
 			fees := c.calculateFees(gasWanted)
--- a/da/celestia/celestia.go
+++ b/da/celestia/celestia.go
@@ -19,6 +19,8 @@ import (
 	"github.com/dymensionxyz/dymint/store"
 	"github.com/dymensionxyz/dymint/types"
 	pb "github.com/dymensionxyz/dymint/types/pb/dymint"
+
+	sdktypes "github.com/cosmos/cosmos-sdk/types"
 )
 
 type CNCClientI interface {
@@ -95,6 +97,19 @@ func (c *DataAvailabilityLayerClient) Init(config []byte, pubsubServer *pubsub.S
 		return err
 	}
 
+	if c.config.GasPrices != "" {
+		if c.config.Fee != 0 {
+			return errors.New("can't set both gas prices and fee")
+		}
+		//Check gas fees parsable
+		_, err = sdktypes.NewDecFromStr(c.config.GasPrices)
+		if err != nil {
+			return err
+		}
+	} else if c.config.Fee == 0 {
+		return errors.New("fee or gas prices must be set")
+	}
+
 	c.pubsubServer = pubsubServer
 	// Set defaults
 	c.txPollingRetryDelay = defaultTxPollingRetryDelay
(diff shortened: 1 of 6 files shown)


da/celestia/celestia.go
```diff
@@ -159,62 +174,63 @@ func (c *DataAvailabilityLayerClient) SubmitBatch(batch *types.Batch) da.ResultS
...
-						return res
-					}
-					return da.ResultSubmitBatch{
-						BaseResult: da.BaseResult{
-							Code:     da.StatusSuccess,
-							Message:  "tx hash: " + txResponse.TxHash,
-							DAHeight: uint64(txResponse.Height),
-						},
-					}
+			estimatedGas := float64(EstimateGas([]int{len(blob)}))
```
omritoptix, line 177: why are we creating an int array of blob size and then re-summing it in `EstimateGas` vs just passing the total size of the blob (i.e uint)?

da/celestia/celestia.go
```diff
@@ -159,62 +174,63 @@ func (c *DataAvailabilityLayerClient) SubmitBatch(batch *types.Batch) da.ResultS
...
+			gasWanted := uint64(estimatedGas * gasAdjustment)
```
omritoptix, line 178: I see we added both gasTolerance and gasAdjustment with factor of 12.
Something seems to me off as when I checking data submission cost the gas would be around the equation. 
For example for 24kb it was `283689utia` (which is just a bit more then equation of cost = 78000 + 8 * num_bytes)

da/celestia/celestia.go
```diff
@@ -159,62 +174,63 @@ func (c *DataAvailabilityLayerClient) SubmitBatch(batch *types.Batch) da.ResultS
...
+			fees := c.calculateFees(gasWanted)
+
+			//SubmitPFB sets an error if the txResponse has error, so we check check the txResponse for error
+			txResponse, err := c.client.SubmitPFB(c.ctx, c.config.NamespaceID, blob, int64(fees), gasWanted)
+			if txResponse == nil {
```
omritoptix, line 183: seems like the only difference between this `if` and the one beneath is the error returned and log.
Possibly can combine them both in the same `if`, e.g `if txResponse == nil || txResponse.Code != 0` and then simply create the relevant error.

da/celestia/celestia.go
```diff
@@ -159,62 +174,63 @@ func (c *DataAvailabilityLayerClient) SubmitBatch(batch *types.Batch) da.ResultS
...
+			if daHeight == 0 {
+				c.logger.Debug("Failed to receive DA batch inclusion result. Waiting for inclusion", "txHash", txResponse.TxHash)
+				daHeight, err = c.waitForTXInclusion(txResponse.TxHash)
+				if err != nil {
+					c.logger.Debug("Failed to receive DA batch inclusion result. Emitting health event and trying again", "error", err)
+					res, err := da.SubmitBatchHealthEventHelper(c.pubsubServer, c.ctx, false, err)
+					if err != nil {
+						return res
+					}
+					time.Sleep(c.submitRetryDelay)
```
omritoptix, line 218: possibly all of this can be exported to a function which gets the error as an argument as this logic repeats in the first 3 if statements.

da/celestia/fees.go
```diff
@@ -0,0 +1,24 @@
...
+	if fees == 0 {
+		fees = int64(c.config.GasPrices * float64(gas))
+	}
+
+	return fees
+}
+
+// EstimateGas estimates the gas required to pay for a set of blobs in a PFB.
+func EstimateGas(blobSizes int) uint64 {
+	totalByteCount := 0
```
omritoptix, line 20: `totalByteCount` seems redundant as we are not summing anything.
Alternative would be to simply change `blobSizes` to `totalBytes` or `blobSize` (we're calculating only for one blob)

omritoptix: Also why not make the EstimateGas part of the `DataAvailabilityLayerClient` struct same as `calculateFees` is?

da/celestia/celestia.go
omritoptix: Latest code was tested with `gasAdjustments=2` and it seems to work. not sure where the discrepancy comes from. can you validate it works for you with `2` and if so update it?

omritoptix (approved):

PR #435, opened 2023-08-08
fix: dymint fails to submit new batches after sl failure

mtsitrin:
# PR Standards

---

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/settlement/dymension/dymension.go
+++ b/settlement/dymension/dymension.go
@@ -90,7 +90,6 @@ type HubClient struct {
 	// channel for getting notified when a batch is accepted by the settlement layer.
 	// only one batch of a specific height can get accepted and we can are currently sending only one batch at a time.
 	// for that reason it's safe to assume that if a batch is accepted, it refers to the last batch we've sent.
-	batchAcceptedCh        chan bool
 	batchRetryAttempts     uint
 	batchRetryDelay        time.Duration
 	batchAcceptanceTimeout time.Duration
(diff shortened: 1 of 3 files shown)


settlement/dymension/dymension.go
```diff
@@ -221,55 +227,59 @@ WaitForBatchAcceptance:
 				utils.SubmitEventOrPanic(d.ctx, d.pubsub, heatlhEventData,
 					map[string][]string{settlement.EventTypeKey: {settlement.EventSettlementHealthStatus}})
 				// Sleep to allow context cancellation to take effect before retrying
+
+				//TODO(mtsitrin): I think it should panic here, if it fails after retries. same as in DA submission
```
omritoptix, line 231: what would be the benefit of that? 

As I see it we're in a constant retry. we have nothing else to do but retry and emit health/unhealthy events according to the situation.

omritoptix (approved):

omritoptix:
@mtsitrin tests are failing

settlement/dymension/dymension.go
mtsitrin: I think we need to define whether we want an endless retries loop, or panic after some threshold
currently it's a mix. it will panic if DA fails for too many retries, but won't for SL.

For sure we can leave it as is, but it's making all the retries management useless (as it's in an endless loop)
The RPC isn't responsive in case of unhealthy, so not sure what the difference is.

PR #438, opened 2023-08-09
fix: avail submission was stuck waiting for finalization

mtsitrin:
# PR Standards

---

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/block/manager.go
+++ b/block/manager.go
@@ -10,7 +10,7 @@ import (
 	"code.cloudfoundry.org/go-diodes"
 
 	"cosmossdk.io/errors"
-	"github.com/avast/retry-go"
+	"github.com/avast/retry-go/v4"
 	cryptocodec "github.com/cosmos/cosmos-sdk/crypto/codec"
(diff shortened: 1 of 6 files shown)


omritoptix (commented):
We only want to panic if the err is from a configuration error (otherwise we keep trying to submit and emit health events).

Seems like you changed the errors returned that they can't be detected as `ErrBroacastConfigError` so this part will never execute.

omritoptix (commented):
We only want to panic if the err is from a configuration error (otherwise we keep trying to submit and emit health events) (line 260)

Seems like you changed the errors returned from `broadcastTx` that they can't be detected as `ErrBroacastConfigError` so this part will never execute.

omritoptix:
we're only panicing for configuration/bugs errors not for network errors. both for SL and DA. 

you can see for example in the SL we're panicing if can't find a key, not passing validation etc.  

We don't panic for DA if we failed to send due to network error. same as we don't panic for SL if we failed to send due to network error.

omritoptix (approved):

PR #527, opened 2023-11-19
fix: checking DA received batches

mtsitrin:
# PR Standards

---

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/block/retriever.go
+++ b/block/retriever.go
@@ -21,7 +21,10 @@ func (m *Manager) RetriveLoop(ctx context.Context) {
 		default:
 			// Get only the latest sync target
 			syncTarget := syncTargetpoller.Next()
-			m.syncUntilTarget(ctx, *(*uint64)(syncTarget))
+			err := m.syncUntilTarget(ctx, *(*uint64)(syncTarget))
+			if err != nil {
+				panic(err)
+			}
 			// Check if after we sync we are synced or a new syncTarget was already set.
 			// If we are synced then signal all goroutines waiting on isSyncedCond.
 			if m.store.Height() >= atomic.LoadUint64(&m.syncTarget) {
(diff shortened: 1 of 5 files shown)


block/retriever.go
```diff
@@ -37,26 +40,36 @@ func (m *Manager) RetriveLoop(ctx context.Context) {
...
 // the actual blocks from the DA.
-func (m *Manager) syncUntilTarget(ctx context.Context, syncTarget uint64) {
+func (m *Manager) syncUntilTarget(ctx context.Context, syncTarget uint64) error {
 	currentHeight := m.store.Height()
 	for currentHeight < syncTarget {
-		m.logger.Info("Syncing until target", "current height", currentHeight, "syncTarget", syncTarget)
-		resultRetrieveBatch, err := m.settlementClient.RetrieveBatch(atomic.LoadUint64(&m.lastState.SLStateIndex) + 1)
+		currStateIdx := atomic.LoadUint64(&m.lastState.SLStateIndex) + 1
+		m.logger.Info("Syncing until target", "height", currentHeight, "state_index", currStateIdx, "syncTarget", syncTarget)
+		settlementBatch, err := m.settlementClient.RetrieveBatch(currStateIdx)
```
omritoptix, line 48: The retry before was done using the for loop - i.e while we couldn't sync until target we keep trying. now on the first error you panic (e.g if the rpc of the hub was down). 
It's fine in essence to panic once an error is reached, but I think we should add a retry mechanism inside the retreiveBatch logic (specifically in the dymension client) as we don't want to panic on the first attempt.

da/celestia/celestia.go
```diff
@@ -269,8 +269,12 @@ func (c *DataAvailabilityLayerClient) RetrieveBatches(dataLayerHeight uint64) da
 		var batch pb.Batch
 		err = proto.Unmarshal(msg, &batch)
 		if err != nil {
-			c.logger.Error("failed to unmarshal batch", "daHeight", dataLayerHeight, "position", i, "error", err)
-			continue
+			return da.ResultRetrieveBatch{
```
omritoptix, line 272: Anyone can write to a celestia da namespace. This means that every one who write a random blob which managed to fit into the same height as the rollapp batch, would crash the process. this is undesirable. in essence, lack of ability to unmarshal the blob doesn't mean that something went wrong.

mtsitrin: ok good to know. will fix

in this case, the original version had a bug as well, as we created batch array with `len(data)`. than it could be only partially filled
than we iterate over the batch array with nil values inside

block/retriever.go
mtsitrin: will fix. in the original the retry would happen on the next state update event?
as the `for loop` will be blocked on `syncTarget := syncTargetpoller.Next()`

omritoptix (approved):

Issue #284, opened 2023-03-12
Full nodes can't catch up on sequencer block gossiping once out of sync


omritoptix:
block gossiping is a mechanism in which full nodes track sequencer blocks optimistically in real time.

the importance of the realtime is to provide with minimal latency for users subscribed to the full node waiting for query results. 

the current problem is that if the full node which optimistically listens to gossiped block, misses a block, it can't track the sequencer anymore and has to wait to get the batch from the DA (which entails much bigger latency).

PR #540, opened 2023-12-20
fix: nodes keep out of sync when missing gossiped block (issue #284)

srene:
# PR Standards

---

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


srene:
Fixing the following issue https://github.com/dymensionxyz/dymint/issues/284. 
The problem is when nodes are disconnected or join the network with some delay, they are unable to receive missing blocks and keep out of sync till next batch submission.
The solution enables that nodes can receive previously missed blocks when disconnected and keep up with the block height.
The solution relies in the following:
- Modification to go-libp2p-pubsub to emit gossip of cached messages to any node when the number of nodes of the network is small (below the degree D), otherwise no gossip (periodic broadcast of previously received messages) is received by nodes.
- Caching of out-of-order received blocks, to apply them later when missing block received.
- Modification of default gossipsub parameters to enable previous blocks caching

omritoptix:
@srene tests fail

srene:
@omri ci problems solved

omritoptix (commented):
IMO applyBlock should only applyBlock and it's being called from the sequencer or full nodes but adding caching logic there is not relevant for some of the callers (like the sequncer). 

You can simply check in the `ApplyBlockCallback` before even calling the applyBlock if it should be cached (by the store height etc) . If it’s correct apply it. If not add to cache. And than every time I get a block, assuming my cache is not empty I can try and call `AttemptApplyCachedBlocks` (or some better name) 

This function should probably be called in 2 situations:
1. once we get a new gossiped block
2. once we get a block from the DA

block/retriever.go
```diff
@@ -50,19 +50,22 @@ func (m *Manager) syncUntilTarget(ctx context.Context, syncTarget uint64) error
 			return err
 		}
 
-		if settlementBatch.StartHeight != currentHeight+1 {
+		//code commented out because it may be the case the start height from the da batch is not exactly the next block from the last received
```
omritoptix, line 53: you can simply delete the entire lines. no need for leaving it with comments.

block/retriever.go
```diff
@@ -50,19 +50,22 @@ func (m *Manager) syncUntilTarget(ctx context.Context, syncTarget uint64) error
...
 		err = m.processNextDABatch(ctx, settlementBatch.MetaData.DA.Height)
 		if err != nil {
 			return err
 		}
 
 		currentHeight = m.store.Height()
-		if currentHeight != settlementBatch.EndHeight {
+
+		//code commented out because it may be the case the current height is higher than da height after applying pending blocks, in case of missing blocks
+		/*if currentHeight != settlementBatch.EndHeight {
```
omritoptix, line 66: you can simply delete the entire lines. no need for leaving it with comments.

block/retriever.go
srene: done

block/retriever.go
srene: done

block/block.go
```diff
@@ -112,6 +113,30 @@ func (m *Manager) applyBlock(ctx context.Context, block *types.Block, commit *ty
 	return nil
 }
 
+func (m *Manager) checkPrevCachedBlocks(ctx context.Context) error {
```
omritoptix, line 116: Probably best to create a mutex on the this function as can be called from both DA code and gossiped blocks callback which may lead to a Race condition.

omritoptix, line 116: as this function is not only checking it's also applying I think the name is confusing. Also when it says "Prev" blocks it can be confusing as it has been previously cached but those blocks can be in the future (i.e not yet applied)

block/manager.go
```diff
@@ -212,13 +217,25 @@ func (m *Manager) healthStatusEventCallback(event pubsub.Message) {
...
+	m.logger.Debug("Received new block event", "eventData", event.Data(), "cachedBlocks", len(m.prevBlock))
 	eventData := event.Data().(p2p.GossipedBlock)
 	block := eventData.Block
 	commit := eventData.Commit
-	err := m.applyBlock(context.Background(), &block, &commit, blockMetaData{source: gossipedBlock})
-	if err != nil {
-		m.logger.Debug("Failed to apply block", "err", err)
+
+	if block.Header.Height != m.store.Height()+1 {
+		// We crashed after the commit and before updating the store height.
```
omritoptix, line 226: This comment is actually not accurate as we can cache blocks which just came out of order regardless to any crash.

block/manager.go
```diff
@@ -212,13 +217,25 @@ func (m *Manager) healthStatusEventCallback(event pubsub.Message) {
...
+		m.prevBlock[block.Header.Height] = &block
```
omritoptix, line 227: We should probably try to apply the cached blocks even if we just applied a block because and not just if the blocks are out of order. for example,  if we got previously block at (store height + 2) and put it into cache  and now we got storeHeight + 1 (which means we won't go into the "if")  we can still apply the cached block and no need to wait for the next out-of-order block in order to try it.. 

I would simply add at the end of the function a check, if the block we have in cache is == m.storeHeight + 1 then run `ApplyCachedBlock`.

block/manager.go
```diff
@@ -212,13 +217,25 @@ func (m *Manager) healthStatusEventCallback(event pubsub.Message) {
...
+		m.prevCommit[block.Header.Height] = &commit
+		m.logger.Debug("Caching block", "block height", block.Header.Height, "store height", m.store.Height())
+		err := m.checkPrevCachedBlocks(context.Background())
+		if err != nil {
+			m.logger.Debug("Error applying previous cached blocks", "err", err)
```
omritoptix, line 232: Im not sure if `Error` is correct here as it’s not necessarily due to error - it’s mostly based on attempting. So maybe “failed” . especially when you log it with a "debug"

block/manager.go
```diff
@@ -212,13 +217,25 @@ func (m *Manager) healthStatusEventCallback(event pubsub.Message) {
...
 	commit := eventData.Commit
-	err := m.applyBlock(context.Background(), &block, &commit, blockMetaData{source: gossipedBlock})
-	if err != nil {
-		m.logger.Debug("Failed to apply block", "err", err)
+
+	if block.Header.Height != m.store.Height()+1 {
+		// We crashed after the commit and before updating the store height.
+		m.prevBlock[block.Header.Height] = &block
+		m.prevCommit[block.Header.Height] = &commit
+		m.logger.Debug("Caching block", "block height", block.Header.Height, "store height", m.store.Height())
```
omritoptix, line 229: IMO makes more send to move the  “caching block” message before the actual caching and not after.

block/manager.go
```diff
@@ -212,13 +217,25 @@ func (m *Manager) healthStatusEventCallback(event pubsub.Message) {
...
+		err := m.checkPrevCachedBlocks(context.Background())
```
omritoptix, line 230: Why context.background() is being passed? seems like it’s not being used.

block/block.go
srene: true. mutex added

block/manager.go
srene: done

block/manager.go
srene: log message changed

block/manager.go
srene: true. having function call in the if condition may need an extra block to apply all cached blocks. it's been changed

block/manager.go
srene: it is actually used when calling applyBlock. there is no context in the manager so  context.background() is used by now

block/block.go
srene: name changed to attemptApplyCachedBlocks

block/manager.go
```diff
@@ -224,10 +224,11 @@ func (m *Manager) applyBlockCallback(event pubsub.Message) {
 	commit := eventData.Commit
 
 	if block.Header.Height != m.store.Height()+1 {
-		m.prevBlock[block.Header.Height] = &block
-		m.prevCommit[block.Header.Height] = &commit
-		m.logger.Debug("Caching block", "block height", block.Header.Height, "store height", m.store.Height())
-
+		if block.Header.Height > m.store.Height() {
```
omritoptix, line 227: prob cleaner to do it in the same line as the previous if.

omritoptix (approved):

PR #544, opened 2023-12-26
feat!: using celestia RPC instead of Rest API

mtsitrin:
# PR Standards

---

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


da/celestia/celestia.go
```diff
@@ -187,32 +202,24 @@ func (c *DataAvailabilityLayerClient) SubmitBatch(batch *types.Batch) da.ResultS
 				continue
 			}
 
-			if txResponse.Code != 0 {
-				c.logger.Error("Failed to submit DA batch. Emitting health event and trying again", "txResponse", txResponse.RawLog, "code", txResponse.Code)
-				res, err := da.SubmitBatchHealthEventHelper(c.pubsubServer, c.ctx, false, errors.New(txResponse.RawLog))
+			if txResponse == nil {
```
omritoptix, line 205: Seems like we can skip this response and the next (txresponse.Code) as they are doing it currently in their new api: https://github.com/celestiaorg/celestia-node/blob/ccf9b56317ae8941f91c5d09f24fe8422bbe64bb/state/core_access.go#L256

da/celestia/celestia.go
```diff
@@ -187,32 +202,24 @@ func (c *DataAvailabilityLayerClient) SubmitBatch(batch *types.Batch) da.ResultS
...
 				continue
 			}
 
-			// Here we assume that if txResponse is not nil and also error is not nil it means that the transaction
-			// was submitted (not necessarily accepted) and we still didn't get a clear status regarding it (e.g timeout).
-			// hence trying to poll for it.
-			daHeight := uint64(txResponse.Height)
-			if daHeight == 0 {
-				c.logger.Debug("Failed to receive DA batch inclusion result. Waiting for inclusion", "txHash", txResponse.TxHash)
-				daHeight, err = c.waitForTXInclusion(txResponse.TxHash)
```
omritoptix, line 206: Why did we remove the inclusion test?
the submitPayForBlock broadcasts the tx in `sync` mode which means it's not necessarily gonna get included in a block.
Also from their docs: https://docs.celestia.org/nodes/transaction-resubmission#transaction-resubmission

mtsitrin: it's actually in block mode and by their docs it will fail after 75 sec, and the caller expected to retry.
this implementation is not relevant IMO

We do plan to get the inclusion proof/commitment, we have separate issue for it.

da/celestia/celestia.go
```diff
@@ -164,20 +168,31 @@ func (c *DataAvailabilityLayerClient) SubmitBatch(batch *types.Batch) da.ResultS
...
+		return da.ResultSubmitBatch{
+			BaseResult: da.BaseResult{
+				Code:    da.StatusError,
+				Message: err.Error(),
+			},
+		}
+	}
+	blobs := []*blob.Blob{blockBlob}
+
+	estimatedGas := DefaultEstimateGas(uint32(len(data)))
```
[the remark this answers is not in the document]
mtsitrin: no, it's the same as in celestia-node

da/celestia/celestia.go
mtsitrin: validating `txResponse != nil` is good practice to avoid possible panics.
i'll remove the code check

da/celestia/celestia.go
```diff
@@ -202,24 +202,16 @@ func (c *DataAvailabilityLayerClient) SubmitBatch(batch *types.Batch) da.ResultS
...
 
+			//double check txResponse is not nil - not supposed to happen
 			if txResponse == nil {
-				c.logger.Error("Failed to submit DA batch. Emitting health event and trying again", "error", "txResponse is nil")
-				res, err := da.SubmitBatchHealthEventHelper(c.pubsubServer, c.ctx, false, errors.New("txResponse is nil"))
-				if err != nil {
-					return res
+				err := errors.New("txResponse is nil")
+				c.logger.Error("Failed to submit DA batch", "error", err)
+				return da.ResultSubmitBatch{
```
omritoptix, line 209: why are we returning here and not retrying and submitting an health event? the node will simply panic.

omritoptix:
tests failing

da/celestia/celestia.go
mtsitrin: well it's not supposed to happen, so IMO panic is the right choice (similar to invariant break)
WDYT

omritoptix: I don't see it as an invariant broken or a bug so I don't think we should panic, especially if retry could help (which is why we added it in the previous version as it was the celestia's network instability/response).

Also if we panic and restart don't see why would it change anything if it depends on the celestia network.

mtsitrin: it's not depend on the network
it's part of the API contract. u don't suppose to get nil resp and nil err

but ok I'll change it

omritoptix (approved):

Issue #552, opened 2024-01-01
Separate logic of aggregator with non-aggregator


mtsitrin:
current issues in the implementation:

- multiple redundant events in case of batch submission
  -- `HubClient` gets events from SL and sends `NewSettlementBatchAccepted`
  -- `stateUpdatesHandler()` gets this event and sends additional event `EventNewBatchAccepted`

- too complex flow for an aggregator to be synced with correct state height
-- sequencer sends a batch, but reiles on the internal events to update the `syncTarget`
-- aggregator can set the `synctarget` on it's own after successful SL submit

design changes:
* aggregator runs `produce` and `submit`
* non-aggregator runs `sync` and `retreive`
* submit loop sets the syncTarget after successful submission (saving 2 internal events)
* single event type for SL batch acceptance

PR #558, opened 2024-01-03
refactor: separate logic of aggregator with non aggregator

mtsitrin:
from the issue (https://github.com/dymensionxyz/dymint/issues/552):

> current issues in the implementation:
> 
> * multiple redundant events in case of batch submission
>   -- `HubClient` gets events from SL and sends `NewSettlementBatchAccepted`
>   -- `stateUpdatesHandler()` gets this event and sends additional event `EventNewBatchAccepted`
> * too complex flow for an aggregator to be synced with correct state height
>   -- sequencer sends a batch, but reiles on the internal events to update the `syncTarget`
>   -- aggregator can set the `synctarget` on it's own after successful SL submit
> 
> design changes:
> 
> * aggregator runs `produce` and `submit`
> * non-aggregator runs `sync` and `retreive`
> * submit loop sets the syncTarget after successful submission (saving 2 internal events)
> * single event type for SL batch acceptance

Code was tested as sequencer, sycning sequencer, and a full node


omritoptix:
tests failed on race condition

omritoptix (dismissed):
beautiful job 💯

omritoptix:
linters fail

omritoptix (approved):

PR #573, opened 2024-01-31
chore: Change default block time and batch submit time in dymint.toml

### LINKED ISSUE #502: Change default block time and batch submit time in dymint.toml
should be similar to what is set in roller currently

deepakgudla:
fix: changed Change default block time and batch submit time in dymint.toml as set in roller currently

ref: closes #502

### DIFF
--- a/config/defaults.go
+++ b/config/defaults.go
@@ -26,9 +26,9 @@ func DefaultConfig(home, chainId string) *NodeConfig {
 			Seeds:         ""},
 		Aggregator: true,
 		BlockManagerConfig: BlockManagerConfig{
-			BlockTime:               200 * time.Millisecond,
+			BlockTime:               200 * time.Microsecond,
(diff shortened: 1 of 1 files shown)


deepakgudla:
Hello @omritoptix , Thank you for letting me know that the PR doesn't match the roller config, will it be fine if I change the block time from micro seconds to milli seconds, does this work? If not,can you please let me know
ps:  I have followed this [file](https://github.com/dymensionxyz/roller/blob/main/test/config/init/goldens/init_without_flags/rollapp/config/dymint.toml#L2-L5)
Thank you

omritoptix:
yea should be 0.2 seconds which is 200 miliseconds (not micro)

deepakgudla:
@omritoptix Thanks for the suggestion, I have updated the changes. If there are any modifications, please let me know
Thank you

omritoptix (dismissed):

omritoptix (approved):

Issue #422, opened 2023-07-24
Add validation for block inclusion when using celestia


omritoptix:
Celestia requires to manually call the block inclusion proof api to prove a block is available. 
Currently we ignore this step and assume it's included.

Blocked by #398

mtsitrin:
@omritoptix can u refer to their documentation for this?
e.g Rollkit has nothing regarding this

omritoptix:
I believe we need to run this 2: 
https://node-rpc-docs.celestia.org/?version=v0.12.2#blob.GetProof
https://node-rpc-docs.celestia.org/?version=v0.12.2#blob.Included

PR #575, opened 2024-02-02
feat: Add data availability validation for Celestia DA #422

srene:
# PR Standards

---

For Author:

- [X]  Targeted PR against correct branch
- [X]  included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [X]  Linked to Github issue with discussion and accepted design
- [X]  Targets only one github issue

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


da/celestia/celestia.go
```diff
@@ -223,36 +232,105 @@ func (c *DataAvailabilityLayerClient) SubmitBatch(batch *types.Batch) da.ResultS
...
+			}
+		}
+	}
+
+}
+
+func (c *DataAvailabilityLayerClient) RetrieveBatches(daMetaData *da.DAMetaData) da.ResultRetrieveBatch {
+
+	//Just for backward compatibility, in case no commitments are sent from the Hub, batch can be retrieved using previous implementation.
+	if daMetaData.Commitments == nil || len(daMetaData.Commitments) == 0 {
```
omritoptix, line 269: This is actually a security hole so not sure we can release it like that.
The big majority of Froopyland don't run full nodes. Once we upgrade them assuming not synced, they will have to sync from a snapshot from the point that state updates must contain commitments.
I suggest removing it.

da/celestia/celestia.go
```diff
@@ -223,36 +232,105 @@ func (c *DataAvailabilityLayerClient) SubmitBatch(batch *types.Batch) da.ResultS
...
+							Code:    da.StatusBlobNotFound,
+							Message: "Blob not found",
+						},
+					}
+				}
+
+				var batch pb.Batch
+				err = proto.Unmarshal(blob.Data, &batch)
+				if err != nil {
+					c.logger.Error("failed to unmarshal block", "daHeight", daMetaData.Height, "error", err)
```
omritoptix, line 303: failed to unmarshal `batch` instead of `block`

block/retriever.go
```diff
@@ -99,19 +99,54 @@ func (m *Manager) processNextDABatch(ctx context.Context, daHeight uint64) error
 	return nil
 }
 
-func (m *Manager) fetchBatch(daHeight uint64) (da.ResultRetrieveBatch, error) {
+func (m *Manager) fetchBatch(daMetaData *da.DAMetaData) (da.ResultRetrieveBatch, error) {
 	var err error
-	batchRes := m.retriever.RetrieveBatches(daHeight)
+	availRes := m.retriever.CheckBatchAvailability(daMetaData)
+
+	switch availRes.Code {
```
omritoptix, line 106: imo leaving the code each status should indiciate a different handling. i.e `success` means on thing, `errorGettingProof` can be another and `ProofInvalid` can be another one but if they have the same handling route than the status code a bit looses the point IMO. If you want to add extra info to each status you can add it in the `message` field.

block/retriever.go
```diff
@@ -99,19 +99,54 @@ func (m *Manager) processNextDABatch(ctx context.Context, daHeight uint64) error
...
+		err = fmt.Errorf("Unable to get proof on height %d for the commitment", daMetaData.Height)
+
+	case da.StatusProofNotMatching:
+		//The proofs are obtained for the commitment, but not matching with the span (index, length) committed to the Hub.
+		err = fmt.Errorf("Span not matching the commitment")
+	case da.StatusError:
+		//There's been an issue validating proofs
+		err = fmt.Errorf("Error validating batch")
+	}
+	if !availRes.DataAvailable {
```
omritoptix, line 119: Also here would change to `availabilityRes` as we have `avail` and it's confusing

block/retriever.go
```diff
@@ -99,19 +99,54 @@ func (m *Manager) processNextDABatch(ctx context.Context, daHeight uint64) error
...
+	}
+	//batchRes.MetaData includes proofs necessary to open disputes with the Hub
+	batchRes := m.retriever.RetrieveBatches(daMetaData)
+	batchRes.MetaData = availRes.MetaData
+
 	switch batchRes.Code {
 	case da.StatusError:
-		err = fmt.Errorf("failed to retrieve batch from height %d: %s", daHeight, batchRes.Message)
+		err = fmt.Errorf("failed to retrieve batch from height %d: %s", daMetaData.Height, batchRes.Message)
 	case da.StatusTimeout:
```
omritoptix, line 136: Same here with the switch case of enums like above

block/retriever.go
```diff
@@ -99,19 +99,54 @@ func (m *Manager) processNextDABatch(ctx context.Context, daHeight uint64) error
...
 	var err error
-	batchRes := m.retriever.RetrieveBatches(daHeight)
+	availRes := m.retriever.CheckBatchAvailability(daMetaData)
+
+	switch availRes.Code {
+	case da.StatusUnableToGetProof:
+		//Is not possible to obtain proofs for the specific commitment in the specific height.
+		//This means there is no matching blob for that commitment
+		err = fmt.Errorf("Unable to get proof on height %d for the commitment", daMetaData.Height)
+
```
omritoptix, line 111: no need for those spaces..

block/retriever.go
omritoptix: Also seems like `DataAvailable`==`statusSuccess` as from it's description. so not sure why both are needed.

block/retriever.go
```diff
@@ -99,19 +99,54 @@ func (m *Manager) processNextDABatch(ctx context.Context, daHeight uint64) error
...
+		//The proofs are obtained for the commitment, but not matching with the span (index, length) committed to the Hub.
+		err = fmt.Errorf("Span not matching the commitment")
+	case da.StatusError:
+		//There's been an issue validating proofs
+		err = fmt.Errorf("Error validating batch")
+	}
+	if !availRes.DataAvailable {
+		err = fmt.Errorf("no batches found on height %d", daMetaData.Height)
+		//There is no point on fetching the data
+		batchRes := da.ResultRetrieveBatch{}
```
omritoptix, line 122: it's better to just insert the fields in the ResultRetrieveBatch while you create it vs adding it after

block/retriever.go
```diff
@@ -99,19 +99,54 @@ func (m *Manager) processNextDABatch(ctx context.Context, daHeight uint64) error
...
+		batchRes.Code = da.StatusError
+		batchRes.Message = "Error validating data"
+		batchRes.MetaData = availRes.MetaData
+		return batchRes, err
```
omritoptix, line 126: I guess if data is not available we need to seperate to 2 cases:
1. not available due to RPC down etc - in this case - should be retried
2. not available due to fraud, which is another case.

Lets not handle this now but lets make sure we have an issue to handle this.

block/retriever.go
```diff
@@ -99,19 +99,54 @@ func (m *Manager) processNextDABatch(ctx context.Context, daHeight uint64) error
...
+
+	}
+	//batchRes.MetaData includes proofs necessary to open disputes with the Hub
+	batchRes := m.retriever.RetrieveBatches(daMetaData)
+	batchRes.MetaData = availRes.MetaData
```
omritoptix, line 131: error should be checked before assignment

block/retriever.go
```diff
@@ -99,19 +99,54 @@ func (m *Manager) processNextDABatch(ctx context.Context, daHeight uint64) error
...
+
 	switch batchRes.Code {
 	case da.StatusError:
-		err = fmt.Errorf("failed to retrieve batch from height %d: %s", daHeight, batchRes.Message)
+		err = fmt.Errorf("failed to retrieve batch from height %d: %s", daMetaData.Height, batchRes.Message)
```
omritoptix, line 135: creation of error based on case is a bit dirty IMO. As I see it:
1. we can either return the original Error
2. we can simply log the message and return original error
3. we can have the `message` field holding the error which can be either a string or an error object which we call Error() on.

settlement/dymension/dymension.go
```diff
@@ -407,8 +407,14 @@ func (d *HubClient) convertBatchToMsgUpdateState(batch *types.Batch, daClient da
 		return nil, err
 	}
 
-	DAMetaData := &settlement.DAMetaData{
-		Height: daResult.DAHeight,
+	height := uint64(0)
```
omritoptix, line 410: why is this necessary?

settlement/dymension/dymension.go
```diff
@@ -407,8 +407,14 @@ func (d *HubClient) convertBatchToMsgUpdateState(batch *types.Batch, daClient da
...
+	if daResult.MetaData != nil {
+		height = daResult.MetaData.Height
```
omritoptix, line 412: why should the daResult.MetData be nil? why are we checking this?

settlement/dymension/dymension.go
```diff
@@ -407,8 +407,14 @@ func (d *HubClient) convertBatchToMsgUpdateState(batch *types.Batch, daClient da
...
+	}
+
+	//TODO (srene): Include extra dataMetaDATA to the Hub for inclusion validation
+	DAMetaData := &da.DAMetaData{
+		Height: height,
```
omritoptix, line 417: Why are we jus not setting the height to the daResult.Height right away like before?

da/celestia/celestia.go
```diff
@@ -180,7 +186,11 @@ func (c *DataAvailabilityLayerClient) GetClientType() da.Client {
 
 // SubmitBatch submits a batch to the DA layer.
 func (c *DataAvailabilityLayerClient) SubmitBatch(batch *types.Batch) da.ResultSubmitBatch {
+
+	var blobs [][]byte
+
 	data, err := batch.MarshalBinary()
+	blobs = [][]byte{data}
```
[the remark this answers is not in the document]
srene: sure. moved

da/celestia/celestia.go
```diff
@@ -189,30 +199,29 @@ func (c *DataAvailabilityLayerClient) SubmitBatch(batch *types.Batch) da.ResultS
...
 	}
-
-	blockBlob, err := blob.NewBlobV0(c.config.NamespaceID.Bytes(), data)
-	if err != nil {
+	if len(data) > DefaultMaxBytes {
 		return da.ResultSubmitBatch{
 			BaseResult: da.BaseResult{
 				Code:    da.StatusError,
-				Message: err.Error(),
+				Message: "Batch size bigger than maximum blob size",
```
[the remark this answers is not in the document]
srene: ok

da/celestia/celestia.go
```diff
@@ -189,30 +199,29 @@ func (c *DataAvailabilityLayerClient) SubmitBatch(batch *types.Batch) da.ResultS
...
 
 	for {
 		select {
 		case <-c.ctx.Done():
 			c.logger.Debug("Context cancelled")
 			return da.ResultSubmitBatch{}
 		default:
-			txResponse, err := c.rpc.SubmitPayForBlob(c.ctx, math.NewInt(fees), gasWanted, blobs)
+			gasPrice := float64(-1)
+			if c.config.GasPrices >= 0 {
```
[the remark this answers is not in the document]
srene: true. moved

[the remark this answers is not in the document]
srene: I think I got it from rollkit example to indicate is not set, but I actually removed gasPrices as input parameter for the submit() function, because it can be get later on from the config anyway, and no need to pass as a parameter.

da/celestia/celestia.go
```diff
@@ -189,30 +199,29 @@ func (c *DataAvailabilityLayerClient) SubmitBatch(batch *types.Batch) da.ResultS
...
+				gasPrice = c.config.GasPrices
+			}
+			//TODO(srene):  Split batch in multiple blobs if necessary if supported
+			height, commitments, blobs, err := c.submit(blobs, gasPrice)
+
+			c.logger.Info("DA batch submitted", "commitments", len(commitments), "blobs", len(blobs))
```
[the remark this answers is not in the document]
srene: moved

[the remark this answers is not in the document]
srene: I also removed returned blobs from submit function, since it was only used for logging

block/retriever.go
srene: done

settlement/dymension/dymension.go
srene: this was to avoid null pointer errors caused by a test returning empty batch result, which should not happen anyway because it will panic before. I adapted the test and removed this

da/avail/avail.go
```diff
@@ -155,11 +155,12 @@ func (c *DataAvailabilityLayerClient) GetClientType() da.Client {
 }
 
 // RetrieveBatch retrieves batch from DataAvailabilityLayerClient instance.
-func (c *DataAvailabilityLayerClient) RetrieveBatches(dataLayerHeight uint64) da.ResultRetrieveBatch {
-	blockHash, err := c.client.GetBlockHash(dataLayerHeight)
+func (c *DataAvailabilityLayerClient) RetrieveBatches(daMetaData *da.DASubmitMetaData) da.ResultRetrieveBatch {
```
mtsitrin, line 158: something off with the naming, we pass `DASubmitMetaData` to *Retrieve* function

da/avail/avail.go
```diff
@@ -398,9 +402,9 @@ func (c *DataAvailabilityLayerClient) broadcastTx(tx []byte) (uint64, error) {
 }
 
 // CheckBatchAvailability checks batch availability in DataAvailabilityLayerClient instance.
-func (c *DataAvailabilityLayerClient) CheckBatchAvailability(dataLayerHeight uint64) da.ResultCheckBatch {
+func (c *DataAvailabilityLayerClient) CheckBatchAvailability(daMetaData *da.DASubmitMetaData) da.ResultCheckBatch {
 	return da.ResultCheckBatch{
-		BaseResult: da.BaseResult{
+		BaseDACheckResult: da.BaseDACheckResult{
```
mtsitrin, line 407: doesn't it break all avail rollapps? they won't be able to retrieve blocks optimisticlly

da/avail/avail.go
srene: I named DASubmitMetadata, the DA metadata result of the submission, and used for retrieving the data.
I named DACheckMetaData, the DA metadata resulting of the availability checks including inclusion proofs for disputes. 
I can change the naming if you think is not clear enough

da/celestia/celestia.go
```diff
@@ -189,30 +196,29 @@ func (c *DataAvailabilityLayerClient) SubmitBatch(batch *types.Batch) da.ResultS
 			},
 		}
 	}
+	blobs = [][]byte{data}
 
-	blockBlob, err := blob.NewBlobV0(c.config.NamespaceID.Bytes(), data)
-	if err != nil {
+	if len(data) > celtypes.DefaultMaxBytes {
```
mtsitrin, line 201: this value should be validated for dymint config as well (we have max bytes for batch there)

da/celestia/celestia.go
```diff
@@ -189,30 +196,29 @@ func (c *DataAvailabilityLayerClient) SubmitBatch(batch *types.Batch) da.ResultS
...
 
 	for {
 		select {
 		case <-c.ctx.Done():
 			c.logger.Debug("Context cancelled")
 			return da.ResultSubmitBatch{}
 		default:
-			txResponse, err := c.rpc.SubmitPayForBlob(c.ctx, math.NewInt(fees), gasWanted, blobs)
+
+			c.logger.Info("Submitting DA batch")
```
mtsitrin, line 217: log this above, as it will be printed for each retry otherwise

da/celestia/celestia.go
```diff
@@ -180,6 +184,9 @@ func (c *DataAvailabilityLayerClient) GetClientType() da.Client {
 
 // SubmitBatch submits a batch to the DA layer.
 func (c *DataAvailabilityLayerClient) SubmitBatch(batch *types.Batch) da.ResultSubmitBatch {
+
+	var blobs [][]byte
```
mtsitrin, line 188: redundent. aleardy initilized by `blobs = [][]byte{data}`

da/celestia/celestia.go
```diff
@@ -223,40 +229,106 @@ func (c *DataAvailabilityLayerClient) SubmitBatch(batch *types.Batch) da.ResultS
...
-			if txResponse == nil {
-				err := errors.New("txResponse is nil")
-				c.logger.Error("Failed to submit DA batch", "error", err)
+			daMetaData := &da.DASubmitMetaData{
+				Height:      height,
+				Commitments: commitments,
+			}
+
+			availabilityResult := c.CheckBatchAvailability(daMetaData)
+			if availabilityResult.Code != da.StatusSuccess || !availabilityResult.DataAvailable {
```
mtsitrin, line 238: `availabilityResult.Code != da.StatusSuccess`
will fail for avail

da/celestia/celestia.go
```diff
@@ -285,10 +357,264 @@ func (c *DataAvailabilityLayerClient) RetrieveBatches(dataLayerHeight uint64) da
...
+	defer cancel()
+	isIncluded, error := c.rpc.Included(ctx, height, c.config.NamespaceID.Bytes(), proof, commitment)
+	return isIncluded, error
+}
+
+func (c *DataAvailabilityLayerClient) getDataAvailabilityHeaders(height uint64) (*header.DataAvailabilityHeader, error) {
+
+	c.logger.Info("Getting Celestia extended headers via RPC call")
+	ctx, cancel := context.WithTimeout(c.ctx, c.txPollingRetryDelay)
+	defer cancel()
```
mtsitrin, line 573: no need for it

da/celestia/celestia.go
```diff
@@ -285,10 +357,264 @@ func (c *DataAvailabilityLayerClient) RetrieveBatches(dataLayerHeight uint64) da
...
+		commitments = append(commitments, commitment)
+	}
+	return blobs, commitments, nil
+}
+
+func (c *DataAvailabilityLayerClient) validateProof(height uint64, commitment da.Commitment, proof *blob.Proof) (bool, error) {
+
+	c.logger.Info("Getting inclusion validation via RPC call")
+	ctx, cancel := context.WithTimeout(c.ctx, c.txPollingRetryDelay)
+	defer cancel()
```
mtsitrin, line 564: no need for it

da/celestia/celestia.go
```diff
@@ -285,10 +357,264 @@ func (c *DataAvailabilityLayerClient) RetrieveBatches(dataLayerHeight uint64) da
...
+				CheckMetaData: DACheckMetaData,
+			},
+		}
+	}
+	DACheckMetaData.Root = dah.Hash()
+	for i, commitment := range daMetaData.Commitments {
+		included := false
+
+		proof, err := c.getProof(daMetaData.Height, commitment)
+		if err != nil || proof == nil {
```
mtsitrin, line 394: err valiue not needed? not logged

da/celestia/celestia.go
```diff
@@ -285,10 +357,264 @@ func (c *DataAvailabilityLayerClient) RetrieveBatches(dataLayerHeight uint64) da
...
+		for i, proof := range nmtProofs {
+			if i == 0 {
+				indexes = append(indexes, proof.Start())
+			}
+			shares += proof.End() - proof.Start()
+		}
+		numShares = append(numShares, shares)
+
+		if daMetaData.Indexes != nil && daMetaData.Lengths != nil {
+			if indexes[i] != daMetaData.Indexes[i] || shares != daMetaData.Lengths[i] {
```
mtsitrin, line 421: how u know that indexes[i] exists?
indexes appended once, i runs over nmtProofs

srene: if there is no index is because there is no proof returned in previous step, and therefore the function exited before.

da/celestia/celestia.go
srene: it's there to force  a timeout in case rpc call does not timeout by itself (which was happening for some problems in celestia)

da/celestia/celestia.go
srene: same here

da/celestia/celestia.go
```diff
@@ -223,40 +229,106 @@ func (c *DataAvailabilityLayerClient) SubmitBatch(batch *types.Batch) da.ResultS
...
+			}
+
+			availabilityResult := c.CheckBatchAvailability(daMetaData)
+			if availabilityResult.Code != da.StatusSuccess || !availabilityResult.DataAvailable {
+				c.logger.Error("Unable to confirm submitted blob availability. Retrying")
 				res, err := da.SubmitBatchHealthEventHelper(c.pubsubServer, c.ctx, false, err)
 				if err != nil {
 					return res
 				}
 				time.Sleep(c.submitRetryDelay)
```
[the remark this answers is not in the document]
srene: this shouldn't happen, because if the submission was successful, there should be valid proofs available for the blob. however, in case it happens, it makes sense to me to resubmit again, disregarding previous submission result, because from rpc calls it is not possible to know whether there was a problem in the submitted blob or in getting the proofs.

da/celestia/celestia.go
```diff
@@ -285,10 +357,264 @@ func (c *DataAvailabilityLayerClient) RetrieveBatches(dataLayerHeight uint64) da
...
+// Submit submits the Blobs to Data Availability layer.
+func (c *DataAvailabilityLayerClient) submit(daBlobs []da.Blob) (uint64, []da.Commitment, error) {
+	blobs, commitments, err := c.blobsAndCommitments(daBlobs)
+	if err != nil {
+		return 0, nil, err
+	}
+
+	options := openrpc.DefaultSubmitOptions()
+
+	if c.config.GasPrices >= 0 {
```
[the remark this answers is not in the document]
srene: ok. check removed

da/celestia/celestia.go
srene: ok. removed

omritoptix (approved):

PR #592, opened 2024-02-29
fix: changing dymint rpc to enable queries for unhealthy nodes

### LINKED ISSUE #576: upon unhealthy keep queries alive
currently the entire node queries return the error. 
Only the tx endpoints should be disabled and the queries should still function. 
we should have an health endpoint just for that.

srene:
# Description

- currently the entire node queries return error when nodes are set to unhealthy state, after receiving an unhealthy event. 
- with the modifications, we enable a new rpc endpoint"/health", to be able to query the health status of the node, but other rpc are still available. 
- all responses attach the health status of the node

Closes #576 
# PR Standards

---

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/rpc/middleware/status.go
+++ b/rpc/middleware/status.go
@@ -1,9 +1,7 @@
 package middleware
 
 import (
-	"io"
 	"net/http"
-	"net/http/httptest"
 	"strconv"
 
 	"github.com/dymensionxyz/dymint/rpc/sharedtypes"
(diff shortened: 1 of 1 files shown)


rpc/middleware/client.go
```diff
@@ -24,7 +25,7 @@ func NewClient(reg Registry, logger log.Logger) *Client {
 // Handle wraps the provided http.Handler with the registered middlewares and returns the final http.Handler.
 func (mc *Client) Handle(h http.Handler) http.Handler {
 	registeredMiddlewares := mc.registry.GetRegistered()
-
+	fmt.Println("handle")
```
mtsitrin, line 28: remove

rpc/middleware/status.go
```diff
@@ -24,29 +25,58 @@ func NewStatusMiddleware(healthStatus *sharedtypes.HealthStatus) *StatusMiddlewa
...
 func (s *StatusMiddleware) Handler(logger log.Logger) HandlerFunc {
 	return func(h http.Handler) http.Handler {
-		return status(s.healthStatus, h, logger)
-	}
-}
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
+			isHealthy, err := s.healthStatus.Get()
+			//in case the endpoint is health we return health response
+			if r.URL.Path == "/health" {
```
mtsitrin, line 32: IMO it should be added on the handler, no need for middlware to handle this URI

mtsitrin:
@srene why we need to `all responses attach the health status of the node`?

rpc/middleware/client.go
srene: done

PR #594, opened 2024-02-29
fix: skip trying to apply blocks from DA with mismatched heights.

omritoptix:
# PR Standards

---

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/block/retriever.go
+++ b/block/retriever.go
@@ -87,6 +87,9 @@ func (m *Manager) processNextDABatch(ctx context.Context, daMetaData *da.DASubmi
 
 	for _, batch := range batchResp.Batches {
 		for i, block := range batch.Blocks {
+			if block.Header.Height != m.store.Height()+1 {
+				continue
+			}
 			err := m.applyBlock(ctx, block, batch.Commits[i], blockMetaData{source: daBlock, daHeight: daMetaData.Height})
 			if err != nil {
 				return err


block/retriever.go
```diff
@@ -87,6 +87,9 @@ func (m *Manager) processNextDABatch(ctx context.Context, daMetaData *da.DASubmi
 
 	for _, batch := range batchResp.Batches {
 		for i, block := range batch.Blocks {
+			if block.Header.Height != m.store.Height()+1 {
```
mtsitrin, line 90: it's already handled in the `applyBlock`

```
if block.Header.Height != m.store.Height()+1 {
		// We crashed after the commit and before updating the store height.
		m.logger.Error("Block not applied. Wrong height", "block height", block.Header.Height, "store height", m.store.Height())
		return nil
	}
```

omritoptix: I know but it's bad handling because it logs an error as the current assumption is that applyBlock expects to get the correct height - as the comment states this error should be logged upon crash recovery.

This was confusing when trying to debug full node not syncing. 

if we get this error logged for every block on the DA which is not in the wrong height it's bad.

mtsitrin: ok now I understand the pr
BTW, why would a wrong block be committed to the DA?

omritoptix: its not a wrong block committed but if for example there is a batch for 100 blocks and you synced until block 20 and failed, next time you'll run your full node, you'll still get the same 100 blocks batch (as you need 20-100) but you'll start applying from block 1 again instead from block 20.

PR #606, opened 2024-03-19
feat!: passing operartor address to the RDK

mtsitrin:
# PR Standards

This PR implmements `dymint` side of the new sequencer registration flow.
The RDK expects dymint to pass 2 pubkeys on the `InitChain` request:
1. The consensus pubkey
2. The operator address pubkey

To get the operator addr, we will have dedicated config in `dymint.toml`
which will be defaulted to the sequencer key used for state updates

**Additional changes:**
* Init chain is called in aggregator mode only
* local priv key is checked against the Proposer registered on the hub
* go version updated

---
For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


block/manager.go
```diff
@@ -315,49 +298,4 @@ func (m *Manager) getLatestBatchFromSL(ctx context.Context) (*settlement.ResultR
 		return resultRetrieveBatch, err
 	}
 	return resultRetrieveBatch, nil
-
-}
-
-// TODO(omritoptix): possible remove this method from the manager
-func updateInitChainState(s *types.State, res *abci.ResponseInitChain, validators []*tmtypes.Validator) {
```
mtsitrin, line 322: moved to executer

danwt:
Is this ready for review @mtsitrin ? Linter and cov is failing

mtsitrin:
> Is this ready for review @mtsitrin ? Linter and cov is failing

linter seems to fail on some unknown env reason. PR ready to review

danwt:
connected to https://github.com/dymensionxyz/dymension-rdk/commit/07ca2a958a619179a2a3a879e2b82fc4fa394d9e

block/initchain.go
```diff
@@ -0,0 +1,93 @@
...
+	cryptocodec "github.com/cosmos/cosmos-sdk/crypto/codec"
+	"github.com/cosmos/cosmos-sdk/crypto/keyring"
+	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
+	"github.com/evmos/evmos/v12/crypto/hd"
+	"github.com/ignite/cli/ignite/pkg/cosmosaccount"
+	tmtypes "github.com/tendermint/tendermint/types"
+)
+
+func (m *Manager) RunInitChain(ctx context.Context) error {
+	//get the proposer's consensus pubkey
```
danwt, line 19:
```suggestion
func (m *Manager) RunInitChain(ctx context.Context) error {

	// We pass an initial val set which is actually the consensus and operator addresses of the proposer.
	// This is a hack to make sure the chain can get both addresses without us needing to change comet signatures:
	//  The RDK will save a sequencer, and delete the extra validator after initChain, so we also delete it here
	//  to keep the state in sync.
	
	//get the proposer's consensus pubkey
```

block/manager.go
```diff
@@ -128,29 +128,6 @@ func NewManager(
...
-		if err != nil {
-			return nil, err
-		}
-
-		updateInitChainState(&s, res, validators)
-		if _, err := store.UpdateState(s, nil); err != nil {
-			return nil, err
-		}
-	}
-
```
danwt, line 153: I guess moving this to Start is just a refactor?

danwt (dismissed):
Looks good

I suggested an explanation comment

One question about terminology:

Does aggregator mode mean that the node is the current active sequencer (= proposer)?

danwt:
Is the linter thing something to do with mocks?

block/manager.go
mtsitrin: also called now only by the aggregator (was in common flow before)

danwt: Ah ok, in terminology, does aggregator = current active sequencer = proposer?

mtsitrin: terminology wise, yes.

initially aggreagator mode was only set by a flag in dymint.toml,
in this PR I've added validation to this assumption
(local key matches the proposer sequencer)

danwt (approved):

srene (approved):

PR #620, opened 2024-03-28
feat: allow dymint to start with genesis height > 0

mtsitrin:
# PR Standards

---

Close #XXX

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


block/block.go
```diff
@@ -17,9 +17,15 @@ import (
 // In case the following doesn't hold true, it means we crashed after the commit and before updating the store height.
 // In that case we'll want to align the store with the app state and continue to the next block.
 func (m *Manager) applyBlock(ctx context.Context, block *types.Block, commit *types.Commit, blockMetaData blockMetaData) error {
-	if block.Header.Height != m.store.Height()+1 {
+	//validate block height
+	if m.lastState.IsGenesis() {
```
omritoptix, line 21: as this check will be done every time and it's likely only relevant once, I think it's better to move it to another place. Also not sure whats the scenario we're checking exactly. comments would be helpful.

block/block.go
```diff
@@ -41,6 +47,11 @@ func (m *Manager) applyBlock(ctx context.Context, block *types.Block, commit *ty
 		return err
 	}
 
+	if m.lastState.IsGenesis() {
```
omritoptix, line 50: also for this. will be probably done once so maybe there is a a better place.
also not sure about the scenario we're checking - comments would be helpful also here.

block/block.go
mtsitrin: when u apply block, u expects it's height to be the expected height.
the expected height is different between genesis and post genesis

PR #671, opened 2024-04-15
fix(rpc): set const ReadHeaderTimeout to dymint RPC server

### LINKED ISSUE #652: HTTP server misconfiguration allows Slowloris DoS attacks

mtsitrin:
* setting client timeout to mitigate HTTP based DoS attacks

Close #652 

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


rpc/server.go
```diff
@@ -176,7 +180,10 @@ func (s *Server) startRPC() error {
 
 func (s *Server) serve(listener net.Listener, handler http.Handler) error {
 	s.Logger.Info("serving HTTP", "listen address", listener.Addr())
-	s.server = http.Server{Handler: handler} //#nosec
+	s.server = http.Server{
+		Handler:           handler,
+		ReadHeaderTimeout: ReadHeaderTimeout,
+	}
```
danwt, line 186: can we also have readtimeout

And use timeouthandler to wrap the handler?

danwt (changes requested):

mtsitrin:
I've added `ReadHeaderTimeout` to mitigate network based slow attack (which can happen before the execution of `timeouthandler`)

danwt:
refactor :green_apple: 
timeout: we still want https://pkg.go.dev/net/http#TimeoutHandler and  server.ReadTimeout right?

	// ReadTimeout is the maximum duration for reading the entire
	// request, including the body. A zero or negative value means
	// there will be no timeout.
	//
	// Because ReadTimeout does not let Handlers make per-request
	// decisions on each request body's acceptable deadline or
	// upload rate, most users will prefer to use
	// ReadHeaderTimeout. It is valid to use them both.
	ReadTimeout time.Duration
?

danwt (approved):

mtsitrin:
AFAICT, we want server with `ReadHeaderTimeout` defined and `http.TimeoutHandler` as the handler.

zale144 (approved):

zale144:
@danwt I think this is about the timeout for reading headers, to prevent someone from sending a ridiculously large header that would take forever to read. But I am not sure if `http.TimeoutHandler` would cover that.

PR #691, opened 2024-04-15
fix(mempool): set pre and post tx check funcs after genesis, not only after first block

danwt:
# PR Standards

---

Close https://github.com/dymensionxyz/dymint/issues/662

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


state/executor.go
```diff
@@ -297,10 +302,12 @@ func (e *BlockExecutor) commit(ctx context.Context, state *types.State, block *t
 
 	maxBytes := state.ConsensusParams.Block.MaxBytes
 	maxGas := state.ConsensusParams.Block.MaxGas
-	err = e.mempool.Update(int64(block.Header.Height), fromDymintTxs(block.Data.Txs), deliverTxs, mempool.PreCheckMaxBytes(maxBytes), mempool.PostCheckMaxGas(maxGas))
+	err = e.mempool.Update(int64(block.Header.Height), fromDymintTxs(block.Data.Txs), deliverTxs)
 	if err != nil {
 		return nil, 0, err
 	}
+	e.mempool.SetPreCheckFn(mempool.PreCheckMaxBytes(maxBytes))
+	e.mempool.SetPostCheckFn(mempool.PostCheckMaxGas(maxGas))
```
zale144, line 311: I might be missing the bigger picture, but from what I can tell, every time the `txmp.preCheck` and `txmp.postCheck` functions are updated, it's always by `mempool.PreCheckMaxBytes` and `mempool.PostCheckMaxGas`, respectively. The only thing that changes is the `maxBytes` and `maxGas` values passed and kept inside the closure.
So if the implementation of those functions is always going to be the same, why not just keep those functions static and simply have `maxBytes` and `maxGas` as attributes of `TxMempool` and set them dynamically on update.

danwt: I guess, we're still in keeping with the original rollkit codebase, and may want to use different functions at some point
But I kinda agree
YAGNI
Anyways I suggest we keep the change small for now. We can alway simplify further later

zale144 (dismissed):

mtsitrin (approved):

PR #698, opened 2024-04-16
fix(p2p): handle default mempool check tx error case

### LINKED ISSUE #661: Validating gossiped transactions may block the libp2p validator queue

zale144:
# PR Standards

---

Close #661 

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/p2p/validator_test.go
+++ b/p2p/validator_test.go
@@ -1,0 +1,103 @@
+package p2p_test
+
+import (
+	"testing"
+
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/stretchr/testify/assert"
+	abci "github.com/tendermint/tendermint/abci/types"
+	"github.com/tendermint/tendermint/libs/log"
+
+	"github.com/tendermint/tendermint/types"
+
+	"github.com/dymensionxyz/dymint/mempool"
+	nodemempool "github.com/dymensionxyz/dymint/node/mempool"
+	"github.com/dymensionxyz/dymint/p2p"
+)
+
+func TestValidator_TxValidator(t *testing.T) {
+	type args struct {
+		mp      mempool.Mempool
+		numMsgs int
+	}
+	tests := []struct {
+		name string
+		args args
+		want bool
+	}{
+		{
+			name: "valid: tx already in cache",
+			args: args{
+				mp:      &mockMP{err: mempool.ErrTxInCache},
+				numMsgs: 3,
+			},
+			want: true,
+		}, {
+			name: "valid: mempool is full",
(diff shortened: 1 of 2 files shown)


p2p/validator_test.go
```diff
@@ -0,0 +1,103 @@
...
+				numMsgs: 3,
+			},
+			want: true,
+		}, {
+			name: "valid: mempool is full",
+			args: args{
+				mp:      &mockMP{err: mempool.ErrMempoolIsFull{}},
+				numMsgs: 3,
+			},
+			want: true,
```
danwt, line 41: I know you didn't write it but why is it valid if the MP is full?

```
// TxValidator creates a pubsub validator that uses the node's mempool to check the
// transaction. If the transaction is valid, then it is added to the mempool.
func (v *Validator) TxValidator
```


>If the transaction is valid, then it is added to the mempool.

danwt (approved):
Nice simplification!
I'm a bit confused about the mempool is full case 😕

p2p/validator_test.go
zale144: I was wondering the same thing. @omritoptix ?

zale144: @danwt btw, the audit claims that if there is an unhandled error type in the switch, the goroutine will block. They were wrong, it won't block, for every new tx a new buffered (1) channel is created.

Which makes me wonder, why was there a channel in the first place? we're just grabbing a value from a lower scope 🤷

danwt: yeah, greaaaaaaat question haha

maybe it used to be necessary and it got refactored and they forgot to remove the channel
or it was never necessary and original author simply made a mistake 🤷

danwt: let's anyway merge your nice fix and address the mempool full thing seperately

mtsitrin:
looks good!

omritoptix (approved):

p2p/validator_test.go
omritoptix:
```
// CheckTx executes a new transaction against the application to determine
// its validity and whether it should be added to the mempool.
```

```
// If the application accepts the transaction and the mempool is full, the
// mempool evicts one or more of the lowest-priority transaction whose priority
// is (strictly) lower than the priority of tx and whose size together exceeds
// the size of tx, and adds tx instead. If no such transactions exist, tx is
// discarded.
```

In that sense check tx verifies validity and not the fact that it can be added to the pool (but rather should). even if it can't be added to the pool it doesn't mean it's invalid.

PR #702, opened 2024-04-17
fix(rpc): broken ws upgrade introduced by using http.TimeoutHandler

zale144:
# PR Standards

---

Close #XXX

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


zale144:
> can we use stdlib instead https://pkg.go.dev/net/http#TimeoutHandler ?

Check the title dude 😄

mtsitrin:
@zale144 can u pls elaborate what was the issue and how this change fix it?
looking for timeouts+wss support I can find various resources:
* https://github.com/zeromicro/go-zero/blob/master/rest/handler/timeouthandler.go
* https://github.com/golang/go/issues/54136

danwt:
Also I guess if the old solution failed at runtime we should have a regression test for it

zale144:
@mtsitrin @danwt I got errors trying to connect with tendermint, because `http.Hijacker` was not implemented with `http.TimeoutHandler`. I would suggest not going too far with this, as long as we have timeouts and able to upgrade to ws, I'm happy. I won't lose any sleep if we don't have both at the same time (ws with timeout). wdyt?

danwt:
^ Makes sense
I think WS usually have their own idle connection lifetime anyway

danwt:
Does it work now?

mtsitrin (approved):

danwt (dismissed):
trusting it works

rpc/server.go
```diff
@@ -200,7 +189,7 @@ func (s *Server) serve(listener net.Listener, handler http.Handler) error {
 	s.Logger.Info("serving HTTP", "listen address", listener.Addr())
 	s.server = http.Server{
 		Handler:           handler,
-		ReadHeaderTimeout: ReadHeaderTimeout,
+		ReadHeaderTimeout: readHeaderTimeout,
 	}
```
danwt, line 193: do we want readtimeout too? it's often recommended

mtsitrin: no need IMO, it quite overlaps with the timeouthandler

omritoptix: If I'm not mistaken most rpc methods don't have timeout defined (just the subscribe).

however I guess we could assume that rpc nodes exposed to the public have timeouts defined in the infra level (e.g on the load balancer/reverse-proxy) etc. 

So practically speaking not sure it's a concern. with that being said there is also no arm in adding it I guess.

omritoptix (dismissed):

PR #714, opened 2024-04-19
fix(da): check DA client type before trying to fetch batch

### LINKED ISSUE #651: full-node should check valid da path before trying to get blob
it should be avoided trying to get blob using da path that does not match the da config in the node (e.g. da path is for another da)  and log error according to that

srene:
# PR Standards

---

Close #651 

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/block/errors.go
+++ b/block/errors.go
@@ -5,4 +5,5 @@ import "errors"
 var (
 	ErrNonRecoverable = errors.New("non recoverable")
 	ErrRecoverable    = errors.New("recoverable")
+	ErrWrongDA        = errors.New("wrong DA")
 )
--- a/block/retriever.go
+++ b/block/retriever.go
@@ -117,11 +117,11 @@ func (m *Manager) ProcessNextDABatch(daMetaData *da.DASubmitMetaData) error {
 
 func (m *Manager) fetchBatch(daMetaData *da.DASubmitMetaData) da.ResultRetrieveBatch {
 	// Check DA client
-	if daMetaData.Client != m.dalc.GetClientType() {
+	if daMetaData.Client != m.DAClient.GetClientType() {
 		return da.ResultRetrieveBatch{
 			BaseResult: da.BaseResult{
 				Code:    da.StatusError,
-				Message: fmt.Sprintf("DA client for the batch does not match node config: DA client batch: %s: DA client config: %s", daMetaData.Client, m.dalc.GetClientType()),
+				Message: fmt.Sprintf("DA client for the batch does not match node config: DA client batch: %s: DA client config: %s", daMetaData.Client, m.DAClient.GetClientType()),
 				Error:   ErrWrongDA,
 			},
 		}
--- a/block/retriever.go
+++ b/block/retriever.go
@@ -86,7 +86,7 @@ func (m *Manager) ProcessNextDABatch(daMetaData *da.DASubmitMetaData) error {
 	m.logger.Debug("trying to retrieve batch from DA", "daHeight", daMetaData.Height)
 	batchResp := m.fetchBatch(daMetaData)
 	if batchResp.Code != da.StatusSuccess {
-		m.logger.Error("Failed fetching batch from DA", batchResp.Message)
+		m.logger.Error("Unable to retrieve batch", batchResp.Message)
 		return batchResp.Error
 	}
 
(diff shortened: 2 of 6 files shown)


block/retriever.go
```diff
@@ -79,6 +79,7 @@ func (m *Manager) processNextDABatch(daMetaData *da.DASubmitMetaData) error {
 	m.logger.Debug("trying to retrieve batch from DA", "daHeight", daMetaData.Height)
 	batchResp := m.fetchBatch(daMetaData)
 	if batchResp.Code != da.StatusSuccess {
+		m.logger.Error("Failed fetching batch from DA", batchResp.Message)
```
omritoptix, line 82: @danwt wouldn't like that u use the "Failed' word in an error logger :) Better remove.

block/retriever.go
```diff
@@ -108,6 +109,17 @@ func (m *Manager) processNextDABatch(daMetaData *da.DASubmitMetaData) error {
 }
 
 func (m *Manager) fetchBatch(daMetaData *da.DASubmitMetaData) da.ResultRetrieveBatch {
+
```
omritoptix, line 112: I'd suggest remove the blank line

srene: removed

block/retriever.go
srene: msg udpated

omritoptix (dismissed):

block/retriever.go
```diff
@@ -108,6 +109,16 @@ func (m *Manager) processNextDABatch(daMetaData *da.DASubmitMetaData) error {
 }
 
 func (m *Manager) fetchBatch(daMetaData *da.DASubmitMetaData) da.ResultRetrieveBatch {
+	// Check DA client
+	if daMetaData.Client != m.dalc.GetClientType() {
```
mtsitrin, line 113: is this the only thing need to be validated on the SL batch? what about namespace?

mtsitrin (dismissed):

block/retriever.go
omritoptix: yea good thinking.

srene: i added a namespace check for celestia, which i think is the only da using namespace id. 
However, i think namespace id configuration needs to be refactored, since there are multiple params in the config for that, and actually it could be directly derived from chain_id without having to pass it to the full-node or having a config param for that

da/celestia/celestia.go
```diff
@@ -399,6 +400,17 @@ func (c *DataAvailabilityLayerClient) retrieveBatchesNoCommitment(dataLayerHeigh
 
 func (c *DataAvailabilityLayerClient) CheckBatchAvailability(daMetaData *da.DASubmitMetaData) da.ResultCheckBatch {
 	var availabilityResult da.ResultCheckBatch
+	//check namespace
+	if !slices.Equal(daMetaData.Namespace, c.config.NamespaceID.Bytes()) {
```
omritoptix, line 404: why isn't this check one on the `fetchBatch`? also may be worth to simply write a `validateBasic` method on the daMetaData after fetching and test it all matches the current configuration

block/errors.go
```diff
@@ -5,4 +5,5 @@ import "errors"
 var (
 	ErrNonRecoverable = errors.New("non recoverable")
 	ErrRecoverable    = errors.New("recoverable")
+	ErrWrongDA        = errors.New("wrong DA")
```
danwt, line 8: can we specify a bit more: wrong in what sense?

danwt: I guess it's included in the context so it's fine

da/errors.go
```diff
@@ -4,19 +4,21 @@ import "errors"
 
 var (
 	// ErrFailedTxBuild is returned when transaction build fails.
-	ErrTxBroadcastConfigError = errors.New("Failed building tx")
+	ErrTxBroadcastConfigError = errors.New("failed building tx")
 	// ErrFailedTxBroadcast is returned when transaction broadcast fails.
-	ErrTxBroadcastNetworkError = errors.New("Failed broadcasting tx")
+	ErrTxBroadcastNetworkError = errors.New("failed broadcasting tx")
 	// ErrTxBroadcastTimeout is returned when transaction broadcast times out.
```
danwt, line 10: dont need failed..

omritoptix (approved):

danwt (changes requested):
How about having a manager method to verify the settlement batch instead?

https://github.com/dymensionxyz/dymint/blob/8f0d22e5c61f40190df043b36888e3684d4bfb0e/block/retriever.go#L47-L51

PR #739, opened 2024-04-30
fix(DA): use expo backoff in retries

danwt:
paritally adress https://github.com/dymensionxyz/dymint/issues/742

Also, make the health event stringer, and fix a typo in 1 log message


srene (approved):
I think the min and max delay of the backoff function could be a parameter in the config, but it can be a todo for another PR.

omritoptix (approved):

omritoptix (approved):

PR #773, opened 2024-05-02
fix(pubsub): only use unbuffered channels

danwt:
Buffered channels are just bug prone because the subscription will get closed if the channel fills up


block/synctarget.go
```diff
@@ -3,27 +3,23 @@ package block
...
+	"github.com/tendermint/tendermint/libs/pubsub"
+
 	"code.cloudfoundry.org/go-diodes"
 	"github.com/dymensionxyz/dymint/settlement"
 )
 
 // SyncTargetLoop is responsible for getting real time updates about batches submission.
 // for non aggregator: updating the sync target which will be used by retrieveLoop to sync until this target.
 func (m *Manager) SyncTargetLoop(ctx context.Context) {
-	m.logger.Info("Started sync target loop")
```
omritoptix, line 13: @danwt why you removed the log message?

danwt: woops, good catch

mtsitrin:
can u elaborate on the need of it?
why **internal** channels needs to be unbuffered
if we receive p2p in higher rate than we can process I'm not sure it will help


Anyway, I think we need to set the internal pubsub as unbuffered as well, otherwise u can still fail on publish
```
// BufferCapacity allows you to specify capacity for the internal server's
// queue. Since the server, given Y subscribers, could only process X messages,
// this option could be used to survive spikes (e.g. high amount of
// transactions during peak hours).
func BufferCapacity(cap int) Option {
	return func(s *Server) {
		if cap > 0 {
			s.cmdsCap = cap
		}
	}
}
```

omritoptix (approved):

mtsitrin:
what about `(d *HubClient) PostBatch` for `dymension` package?

danwt:
Thanks, fixed @mtsitrin

danwt:
I think it's worth mentioning that when you use an unbuffered channel, it's not the producer thread that gets blocked if no-one receives

It's actually the internal pub sub thread that gets blocked
https://github.com/dymensionxyz/cometbft/blob/main-dym/libs/pubsub/pubsub.go#L321-L342
so it will manifest as the pub sub system freezing

omritoptix:
@danwt following your comment about pub-sub system freeze, I think we should avoid than using unbuffered but using buffered with reasonable buffer size. wdyt

danwt:
> @danwt following your comment about pub-sub system freeze, I think we should avoid than using unbuffered but using buffered with reasonable buffer size. wdyt

I'm not a huge fan from purity perspective because you just kick the problem down the road
But practically speaking, I think it would be ok for us
But, I think we should be very clear what we do when the limit is reached (subscription cancelled) - maybe it's best to panic the program in this case
(Btw we could potentially implement `recover` to check if the panic is from subscription cancelled, and reboot the whole program)

danwt:
gonna close this pr until we decide on an approach

danwt:
see https://github.com/dymensionxyz/dymint/issues/848

Issue #803, opened 2024-05-06
applying cached blocks received from p2p is slow


srene:
we need to investigate if there is anything blocking in the block application process that makes it slow

PR #818, opened 2024-05-07
fix(p2p): avoid receiving duplicated blocks

### LINKED ISSUE #795: discard duplicated packets received via gossip
it seems nodes can receive duplicated blocks. we need to investigate why, and avoid calling onNewGossipedBlock for duplicated

srene:
# PR Standards

This PR is aimed at preventing the reception of duplicated blocks from p2p network, by updating Gossipsub router configuration. It also prevents adding blocks to cache when already received. This improves performance and syncing times, because receiving too many packets from p2p may block applying blocks received from DA.

---

Close #795 #803 

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


block/manager.go
```diff
@@ -240,21 +240,26 @@ func (m *Manager) onNodeHealthStatus(event pubsub.Message) {
 // TODO: move to gossip.go
 // onNewGossippedBlock will take a block and apply it
 func (m *Manager) onNewGossipedBlock(event pubsub.Message) {
-	m.retrieverMutex.Lock() // needed to protect blockCache access
-	m.logger.Debug("Received new block via gossip", "n cachedBlocks", len(m.blockCache))
 	eventData := event.Data().(p2p.GossipedBlock)
 	block := eventData.Block
 	commit := eventData.Commit
 
+	_, found := m.blockCache[block.Header.Height]
```
mtsitrin, line 247: this also needs mutex access
maybe move right above
`m.blockCache[block.Header.Height] =`

mtsitrin (changes requested):

block/manager.go
srene: moved

block/manager.go
```diff
@@ -237,21 +237,28 @@ func (m *Manager) onNodeHealthStatus(event pubsub.Message) {
...
 	commit := eventData.Commit
 	m.logger.Debug("Received new block via gossip", "height", block.Header.Height, "n cachedBlocks", len(m.blockCache))
+	m.retrieverMutex.Lock() // needed to protect blockCache access
+	_, found := m.blockCache[block.Header.Height]
+	if found {
+		m.retrieverMutex.Unlock()
+		return
+	}
+
+	m.logger.Debug("Received new block via gossip", "block height", block.Header.Height, "store height", m.Store.Height(), "n cachedBlocks", len(m.blockCache))
```
danwt, line 251: almost duplicate line of above line 243

danwt (changes requested):
not sure if I understood what is the point of this pr?
There was too much demand for the mutex before?

block/manager.go
srene: true, it's been re-added by mistake. removed

srene:
> not sure if I understood what is the point of this pr? There was too much demand for the mutex before?

it avoids receiving extra duplicates from gossipsub protocol by passing extra options in p2p/client.go, this reduces the demand of the mutex. also it avoids adding already received blocks to the cache, in case it happens.

p2p/client.go
```diff
@@ -320,9 +320,9 @@ func (c *Client) tryConnect(ctx context.Context, peer peer.AddrInfo) {
 func (c *Client) setupGossiping(ctx context.Context) error {
 	pubsub.GossipSubHistoryGossip = c.conf.GossipCacheSize
 	pubsub.GossipSubHistoryLength = c.conf.GossipCacheSize
-	pubsub.GossipSubMaxIHaveMessages = c.conf.GossipCacheSize
 
-	ps, err := pubsub.NewGossipSub(ctx, c.Host)
+	//We add WithSeenMessagesTTL (with 1 year time) option to avoid ever requesting already seen blocks
+	ps, err := pubsub.NewGossipSub(ctx, c.Host, pubsub.WithSeenMessagesTTL(8760*time.Hour))
```
mtsitrin, line 325: can u explain this 1 year time? previously u had 
`time.Duration(int(c.blockTime.Nanoseconds()*int64(c.conf.GossipCacheSize))))`

srene: i removed the ttl option and replaced gossipsub version with a new version where seen packets are not expiring

block/manager.go
```diff
@@ -237,21 +237,27 @@ func (m *Manager) onNodeHealthStatus(event pubsub.Message) {
...
 	eventData := event.Data().(p2p.GossipedBlock)
 	block := eventData.Block
 	commit := eventData.Commit
-	m.logger.Debug("Received new block via gossip", "height", block.Header.Height, "n cachedBlocks", len(m.blockCache))
+	m.retrieverMutex.Lock() // needed to protect blockCache access
+	_, found := m.blockCache[block.Header.Height]
+	if found {
+		m.retrieverMutex.Unlock()
+		return
+	}
```
danwt, line 248: can you add a comment here explaining why we return early?
Like your explanation here https://github.com/dymensionxyz/dymint/pull/818#issuecomment-2103415147
Because otherwise, to a new reader, it probably looks like a useless branch

danwt (changes requested):

block/manager.go
srene: we return early if the block is already received, since it is not necessary to cache again the same block (and therefore avoid attempting applying blocks if actually no new block is received).

block/manager.go
```diff
@@ -237,21 +237,28 @@ func (m *Manager) onNodeHealthStatus(event pubsub.Message) {
...
 // onNewGossippedBlock will take a block and apply it
 func (m *Manager) onNewGossipedBlock(event pubsub.Message) {
-	m.retrieverMutex.Lock() // needed to protect blockCache access
 	eventData := event.Data().(p2p.GossipedBlock)
 	block := eventData.Block
 	commit := eventData.Commit
-	m.logger.Debug("Received new block via gossip", "height", block.Header.Height, "n cachedBlocks", len(m.blockCache))
+	m.retrieverMutex.Lock() // needed to protect blockCache access
+	_, found := m.blockCache[block.Header.Height]
+	//we return early if the block is already received, since it is not necessary to cache again the same block (and therefore it is not necessary to call attemptApplyCachedBlocks() if actually no new block is received).
```
danwt, line 245:
```suggestion
	// It is not strictly necessary to return early, for correctness, but doing so helps us avoid mutex pressure and unnecessary repeated attempts to apply cached blocks
```

danwt (approved):
nice!

omritoptix (approved):

omritoptix:
@mtsitrin can we merge?

PR #830, opened 2024-05-09
patch to mtsitrin/324: remove validate batch

danwt:
The validate is unnecessary because it's valid by construction


mtsitrin:
this validation found few bugs already (e.g single block which is too big to fit a batch)
u can add the validation check to the creation function

danwt:
I think it's literally impossible by construction now, for this validation to find anything, so it's not needed.
unless I'm making a mistake in reading the code

danwt:
superseded by https://github.com/dymensionxyz/dymint/pull/833

PR #832, opened 2024-05-09
patch to mtsitrin/324: inline sl submit useless wrapper func

danwt:
removes a useless wrapper

PR #833, opened 2024-05-09
patch to mtsitrin/324: simplify block creation

danwt:
replaces https://github.com/dymensionxyz/dymint/pull/832, https://github.com/dymensionxyz/dymint/pull/830

### DIFF
--- a/block/submit.go
+++ b/block/submit.go
@@ -128,7 +128,7 @@ func (m *Manager) HandleSubmissionTrigger() error {
 
 	resultSubmitToDA := m.DAClient.SubmitBatch(nextBatch)
 
-	if resultSubmitToDA.Code == da.StatusSuccess {
+	if resultSubmitToDA.Code != da.StatusSuccess {
 		return fmt.Errorf("submit next batch to da: %s", resultSubmitToDA.Message)
 	}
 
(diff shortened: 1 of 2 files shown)


block/submit.go
```diff
@@ -126,22 +126,19 @@ func (m *Manager) HandleSubmissionTrigger() error {
 		return fmt.Errorf("create next batch to submit: %w", err)
 	}
 
-	if err := m.ValidateBatch(nextBatch); err != nil {
-		return fmt.Errorf("validate batch: %w", err)
-	}
-
 	resultSubmitToDA, err := m.submitNextBatchToDA(nextBatch)
```
omritoptix, line 129: why not do here the same you did with the SL? ie calling the `m.DAClient.SubmitBatch` directly?

block/submit.go
```diff
@@ -217,10 +187,15 @@ func (m *Manager) CreateNextBatchToSubmit(startHeight uint64, endHeightInclusive
 			// Remove the last block and commit from the batch
 			batch.Blocks = batch.Blocks[:len(batch.Blocks)-1]
 			batch.Commits = batch.Commits[:len(batch.Commits)-1]
+
+			if height == startHeight {
+				return nil, fmt.Errorf("batch size exceeds max size: %d", totalSize)
```
omritoptix, line 192: in this case should probably say "block size exceeds max batch size" as its only one block.

block/submit.go
danwt: done

block/submit.go
danwt: done

omritoptix (approved):

Issue #799, opened 2024-05-06
celestia da nil pointer error


srene:
goroutine 62 [running]:
github.com/dymensionxyz/dymint/da/celestia.(*DataAvailabilityLayerClient).getDataAvailabilityHeaders(0x14000368c30, 0x14d1e6)
	/Users/sergi/workspace/dymension/dymint/da/celestia/celestia.go:646 +0x88
github.com/dymensionxyz/dymint/da/celestia.(*DataAvailabilityLayerClient).checkBatchAvailability(0x14000368c30, 0x1401d3ff960)
	/Users/sergi/workspace/dymension/dymint/da/celestia/celestia.go:454 +0xdc
github.com/dymensionxyz/dymint/da/celestia.(*DataAvailabilityLayerClient).CheckBatchAvailability.func1()
	/Users/sergi/workspace/dymension/dymint/da/celestia/celestia.go:426 +0x34
github.com/avast/retry-go/v4.Do.func1()
	/Users/sergi/go/pkg/mod/github.com/avast/retry-go/v4@v4.5.0/retry.go:115 +0x24
github.com/avast/retry-go/v4.DoWithData[...](0x1408dd91718?, {0x1408dd91810, 0x3, 0x7a})
	/Users/sergi/go/pkg/mod/github.com/avast/retry-go/v4@v4.5.0/retry.go:179 +0x564
github.com/avast/retry-go/v4.Do(0x0?, {0x1408dd91810?, 0x1408dd91778?, 0x10261c00c?})
	/Users/sergi/go/pkg/mod/github.com/avast/retry-go/v4@v4.5.0/retry.go:118 +0x4c
github.com/dymensionxyz/dymint/da/celestia.(*DataAvailabilityLayerClient).CheckBatchAvailability(0x14000368c30, 0x1401d3ff960)
	/Users/sergi/workspace/dymension/dymint/da/celestia/celestia.go:425 +0x110
github.com/dymensionxyz/dymint/block.(*Manager).fetchBatch(0x140014def08, 0x1401d3ff960)
	/Users/sergi/workspace/dymension/dymint/block/retriever.go:139 +0x184
github.com/dymensionxyz/dymint/block.(*Manager).ProcessNextDABatch(0x140014def08, 0x1401d3ff960)
	/Users/sergi/workspace/dymension/dymint/block/retriever.go:96 +0xd0
[...]

Issue #770, opened 2024-05-02
seg fault 


danwt:
```
Getting Celestia extended headers via RPC call[]  module=celestia
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x28 pc=0x106ff51a8]

goroutine 70 [running]:
github.com/dymensionxyz/dymint/da/celestia.(*DataAvailabilityLayerClient).getDataAvailabilityHeaders(0x14000bfe0f0, 0x149bd5)
	/Users/danwt/Documents/dym/d-dymint/da/celestia/celestia.go:641 +0x88
```
I think rpc can be nil here https://github.com/dymensionxyz/dymint/blob/61630eb458197abe2440a81426210000dff25d40/da/celestia/celestia.go#L641

srene:
bug fixed in https://github.com/dymensionxyz/dymint/pull/857

PR #857, opened 2024-05-13
fix(da): full-nodes can sync from p2p while DA light client is down or out of sync

### LINKED ISSUE #856: fix(da): full-nodes should not wait for da to start syncing
if da light client is down, or out of sync, a full-node will not start, even though it could start syncing and caching packets from p2p

srene:
# PR Standards

The main objective of this PR is the startup process for Celestia DA for full-nodes. With the modifications, full nodes will not wait for a DA to be in sync (or up) and it will do the syncing process in the background, while the full-node can start receiving and caching gossiped blocks from P2P. It also solves  a nil pointer error that was happening when DA light finished syncing and a full-node was started when was not still in sync. Without the modifications of this PR, a node could start receiving gossiped blocks, but everything received while DA was syncing was disregarded because the block manager was not started, waiting for the celestia light node to finish the syncing process.

---

Close #856 , #799 , https://github.com/dymensionxyz/dymint/issues/770

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/block/manager.go
+++ b/block/manager.go
@@ -160,9 +160,15 @@ func (m *Manager) Start(ctx context.Context) error {
 
 	if isSequencer {
 		// TODO: populate the accumulatedSize on startup
+		err = m.syncBlockManager()
+		if err != nil {
+			return fmt.Errorf("sync block manager: %w", err)
+		}
 		go m.ProduceBlockLoop(ctx)
 		go m.SubmitLoop(ctx)
 	} else {
+		go uevent.MustSubscribe(ctx, m.Pubsub, "applyGossipedBlocksLoop", p2p.EventQueryNewNewGossipedBlock, m.onNewGossipedBlock, m.logger)
+
 		go m.RetrieveLoop(ctx)
 		go m.SyncToTargetHeightLoop(ctx)
 	}
--- a/block/manager.go
+++ b/block/manager.go
@@ -11,8 +11,6 @@ import (
 	"github.com/dymensionxyz/dymint/gerr"
 	"github.com/dymensionxyz/dymint/store"
 
-	uevent "github.com/dymensionxyz/dymint/utils/event"
-
 	"code.cloudfoundry.org/go-diodes"
 
 	"github.com/dymensionxyz/dymint/p2p"
@@ -149,15 +147,21 @@ func (m *Manager) Start(ctx context.Context) error {
 		}
 	}
 
+	//Fullnode loop can start before syncing from DA
 	if !isSequencer {
-		go uevent.MustSubscribe(ctx, m.Pubsub, "applyGossipedBlocksLoop", p2p.EventQueryNewNewGossipedBlock, m.onNewGossipedBlock, m.logger)
+		m.startNonSequencerLoop(ctx)
 	}
 
+	// TODO: populate the accumulatedSize on startup
+	//Wait till DA is up and running
+	<-m.DAClient.Started()
+	//Start syncing from DA
 	err = m.syncBlockManager()
 	if err != nil {
 		return fmt.Errorf("sync block manager: %w", err)
 	}
 
+	//Aggregator producer loop is started only when finished syncing from DA
 	if isSequencer {
 		// TODO: populate the accumulatedSize on startup
 		//Wait till DA is up and running
@@ -173,7 +177,6 @@ func (m *Manager) Start(ctx context.Context) error {
 		go m.RetrieveLoop(ctx)
 		go m.SyncToTargetHeightLoop(ctx)
 	}
-
 	return nil
 }
 
--- a/block/manager.go
+++ b/block/manager.go
@@ -169,10 +169,6 @@ func (m *Manager) Start(ctx context.Context) error {
 		go m.ProduceBlockLoop(ctx)
 		go m.SubmitLoop(ctx)
 	} else {
-		fmt.Println("waiting")
-		<-m.DAClient.Started()
-		fmt.Println("started")
-
 		go uevent.MustSubscribe(ctx, m.Pubsub, "applyGossipedBlocksLoop", p2p.EventQueryNewNewGossipedBlock, m.onNewGossipedBlock, m.logger)
 		go m.RetrieveLoop(ctx)
 		go m.SyncToTargetHeightLoop(ctx)
--- a/block/manager.go
+++ b/block/manager.go
@@ -170,7 +170,6 @@ func (m *Manager) Start(ctx context.Context) error {
 		go m.SubmitLoop(ctx)
 	} else {
 		go uevent.MustSubscribe(ctx, m.Pubsub, "applyGossipedBlocksLoop", p2p.EventQueryNewNewGossipedBlock, m.onNewGossipedBlock, m.logger)
-
 		go m.RetrieveLoop(ctx)
 		go m.SyncToTargetHeightLoop(ctx)
 	}
--- a/block/manager.go
+++ b/block/manager.go
@@ -148,7 +148,7 @@ func (m *Manager) Start(ctx context.Context) error {
 		}
 	}
 
-	//Fullnode loop can start before syncing from DA
+	// Fullnode loop can start before syncing from DA
 	if !isSequencer {
 		go uevent.MustSubscribe(ctx, m.Pubsub, "applyGossipedBlocksLoop", p2p.EventQueryNewNewGossipedBlock, m.onNewGossipedBlock, m.logger)
 		go m.RetrieveLoop(ctx)
@@ -156,15 +156,15 @@ func (m *Manager) Start(ctx context.Context) error {
 	}
 
 	// TODO: populate the accumulatedSize on startup
-	//Wait till DA is up and running
+	// Wait till DA is up and running
 	<-m.DAClient.Started()
-	//Start syncing from DA
+	// Start syncing from DA
 	err = m.syncBlockManager()
 	if err != nil {
 		return fmt.Errorf("sync block manager: %w", err)
 	}
 
-	//Aggregator producer loop is started only when finished syncing from DA
+	// Aggregator producer loop is started only when finished syncing from DA
 	if isSequencer {
 		// TODO: populate the accumulatedSize on startup
 		//Wait till DA is up and running
--- a/block/manager.go
+++ b/block/manager.go
@@ -160,7 +160,6 @@ func (m *Manager) Start(ctx context.Context) error {
 		return fmt.Errorf("sync block manager: %w", err)
 	}
 
-	// Sequencer producer loop and full-node retrieve loop is started only when finished syncing from DA
 	if isSequencer {
 		// Sequencer must wait till DA is synced to start submitting blobs
 		<-m.DAClient.Synced()
--- a/block/manager.go
+++ b/block/manager.go
@@ -169,6 +169,10 @@ func (m *Manager) Start(ctx context.Context) error {
 		go m.ProduceBlockLoop(ctx)
 		go m.SubmitLoop(ctx)
 	} else {
+		fmt.Println("waiting")
+		<-m.DAClient.Started()
+		fmt.Println("started")
+
 		go uevent.MustSubscribe(ctx, m.Pubsub, "applyGossipedBlocksLoop", p2p.EventQueryNewNewGossipedBlock, m.onNewGossipedBlock, m.logger)
 		go m.RetrieveLoop(ctx)
 		go m.SyncToTargetHeightLoop(ctx)
--- a/block/manager.go
+++ b/block/manager.go
@@ -12,6 +12,7 @@ import (
 	"github.com/dymensionxyz/dymint/store"
 
 	"code.cloudfoundry.org/go-diodes"
+	uevent "github.com/dymensionxyz/dymint/utils/event"
 
 	"github.com/dymensionxyz/dymint/p2p"
 	"github.com/libp2p/go-libp2p/core/crypto"
@@ -149,7 +150,9 @@ func (m *Manager) Start(ctx context.Context) error {
 
 	//Fullnode loop can start before syncing from DA
 	if !isSequencer {
-		m.startNonSequencerLoop(ctx)
+		go uevent.MustSubscribe(ctx, m.Pubsub, "applyGossipedBlocksLoop", p2p.EventQueryNewNewGossipedBlock, m.onNewGossipedBlock, m.logger)
+		go m.RetrieveLoop(ctx)
+		go m.SyncTargetLoop(ctx)
 	}
 
 	// TODO: populate the accumulatedSize on startup
--- a/block/manager.go
+++ b/block/manager.go
@@ -148,10 +148,7 @@ func (m *Manager) Start(ctx context.Context) error {
 		}
 	}
 
-	if isSequencer {
-		// Sequencer must wait till DA is synced
-		<-m.DAClient.Synced()
-	} else {
+	if !isSequencer {
 		// Fullnode loop can start before syncing from DA
 		go uevent.MustSubscribe(ctx, m.Pubsub, "applyGossipedBlocksLoop", p2p.EventQueryNewNewGossipedBlock, m.onNewGossipedBlock, m.logger)
 	}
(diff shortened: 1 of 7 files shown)


da/celestia/celestia.go
```diff
@@ -138,39 +139,55 @@ func (c *DataAvailabilityLayerClient) Start() (err error) {
...
-		c.logger.Info("waiting for celestia-node to finish syncing", "height", state.Height, "target", state.ToHeight)
-
-		done := make(chan error, 1)
-		go func() {
-			done <- rpc.Header.SyncWait(c.ctx)
-		}()
-
-		ticker := time.NewTicker(1 * time.Minute)
-		defer ticker.Stop()
+	go retry.Do(
```
mtsitrin, line 142: I think simple for loop, with `state.Finished()` as break condition is more clearer.
IMO we shouldn't retry on errors for those queries, as it's an init process and errors should be popped up

mtsitrin (changes requested):

block/manager.go
```diff
@@ -218,15 +218,21 @@ func (m *Manager) onNewGossipedBlock(event pubsub.Message) {
...
 	block := eventData.Block
 	commit := eventData.Commit
-	m.logger.Debug("Received new block via gossip", "height", block.Header.Height, "n cachedBlocks", len(m.blockCache))
+
+	_, found := m.blockCache[block.Header.Height]
+	if found || block.Header.Height < m.Store.NextHeight() {
+		m.retrieverMutex.Unlock()
+		return
+	}
+	m.logger.Info("Received new block via gossip", "height", block.Header.Height, "n cachedBlocks", len(m.blockCache))
```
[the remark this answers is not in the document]
srene: changed

block/manager.go
```diff
@@ -218,15 +218,21 @@ func (m *Manager) onNewGossipedBlock(event pubsub.Message) {
...
 
 	nextHeight := m.Store.NextHeight()
 	if block.Header.Height >= nextHeight {
 		m.blockCache[block.Header.Height] = CachedBlock{
 			Block:  &block,
 			Commit: &commit,
 		}
-		m.logger.Debug("caching block", "block height", block.Header.Height, "store height", m.Store.Height())
+		m.logger.Info("caching block", "block height", block.Header.Height, "store height", m.Store.Height())
```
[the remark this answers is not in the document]
srene: changed

da/celestia/celestia.go
srene: a simple for loop will not start a celestia syncing process in background. the idea of this pr is that only sequencers must wait for syncing and full nodes can start receiving blocks without waiting for it. i think in this case, retry is necessary in case of any error or disconnection with da light happens, to continue with the process, without having to panic.

danwt:
conflicts @srene

danwt:
btw, solves https://github.com/dymensionxyz/dymint/issues/770?

srene:
> btw, solves #770?

yes, the problem was coming from da stuck in the starting process when celestia was syncing meanwhile the node was starting to get batches. I opened a new issue for that, so i think we can close 770

da/celestia/celestia.go
```diff
@@ -138,43 +139,60 @@ func (c *DataAvailabilityLayerClient) Start() (err error) {
...
+					ticker := time.NewTicker(1 * time.Minute)
+					defer ticker.Stop()
+
+					for {
+						select {
+						case err := <-done:
+							if err != nil {
+								return err
+							}
+							return nil
```
danwt, line 168: return err

da/celestia/celestia.go
```diff
@@ -138,43 +139,60 @@ func (c *DataAvailabilityLayerClient) Start() (err error) {
...
 					return err
 				}
-				c.logger.Info("celestia-node still syncing", "height", state.Height, "target", state.ToHeight)
-			}
-		}
-	}
 
-	c.logger.Info("celestia-node is synced", "height", state.ToHeight)
+				if !state.Finished() {
+					c.logger.Info("waiting for celestia-node to finish syncing", "height", state.Height, "target", state.ToHeight)
```
danwt, line 152: seems like a redundant log given the log on 174

da/celestia/celestia.go
```diff
@@ -138,43 +139,60 @@ func (c *DataAvailabilityLayerClient) Start() (err error) {
...
+							return nil
+						case <-ticker.C:
+							state, err := rpc.Header.SyncState(c.ctx)
+							if err != nil {
+								return err
+							}
+							c.logger.Info("celestia-node still syncing", "height", state.Height, "target", state.ToHeight)
+						}
+					}
+				}
```
danwt, line 177: seems this code could be simpler, and only have one line which calls SyncState
which would remove a redundant path

danwt:
```
// Start prepares DataAvailabilityLayerClient to work.
func (c *DataAvailabilityLayerClient) Start() (err error) {
	c.logger.Info("starting Celestia Data Availability Layer Client")

	// other client has already been set
	if c.rpc != nil {
		c.logger.Debug("celestia-node client already set")
		return nil
	}

	var rpc *openrpc.Client
	rpc, err = openrpc.NewClient(c.ctx, c.config.BaseURL, c.config.AuthToken)
	if err != nil {
		return err
	}

	go func() {
		sync := func() error {
			done := make(chan error, 1)
			go func() {
				done <- rpc.Header.SyncWait(c.ctx)
			}()
			ticker := time.NewTicker(1 * time.Minute)
			defer ticker.Stop()
			for {
				select {
				case err := <-done:
					return err
				case <-ticker.C:
					state, err := rpc.Header.SyncState(c.ctx)
					if err != nil {
						return err
					}
					if !state.Finished() {
						c.logger.Info("celestia-node still syncing", "height", state.Height, "target", state.ToHeight)
						continue
					}
					goto syncIsFinished
				}
			}

		syncIsFinished:
			c.rpc = NewOpenRPC(rpc)
			c.started <- struct{}{}
			c.logger.Info("celestia-node is synced", "height", state.ToHeight)

			return nil
		}

		err := retry.Do(sync,
			retry.Attempts(0), // try forever
			retry.Delay(10*time.Second),
			retry.LastErrorOnly(true),
			retry.DelayType(retry.FixedDelay),
		)
		if err != nil {
			c.logger.Error("Waiting for Celestia data availability client to sync", "err", err)
		}
	}()
	return 
}
```

does it work?

danwt (changes requested):

da/celestia/celestia.go
```diff
@@ -138,43 +139,60 @@ func (c *DataAvailabilityLayerClient) Start() (err error) {
...
-			select {
-			case err := <-done:
-				if err != nil {
-					return err
-				}
-				return nil
-			case <-ticker.C:
+	go func() {
+		err := retry.Do(
+			func() error {
```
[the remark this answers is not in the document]
srene: the start is called here: https://github.com/dymensionxyz/dymint/blob/73aae62f30acdc56816ad65550438d8830ac83db/node/node.go#L259

and i think the rpc can be not nil if already set as an option. i guess only used for testing to use with a mock rpc

da/celestia/celestia.go
```diff
@@ -138,43 +139,60 @@ func (c *DataAvailabilityLayerClient) Start() (err error) {
...
 				state, err := rpc.Header.SyncState(c.ctx)
 				if err != nil {
+					c.logger.Error("Failed to start Celestia DA.", "Error", err)
 					return err
```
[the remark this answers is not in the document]
srene: if not logged, the retry func will swallow the error

da/celestia/celestia.go
srene: i refactored the function using yours as a base

block/manager.go
```diff
@@ -144,24 +143,27 @@ func (m *Manager) Start(ctx context.Context) error {
 		}
 	}
 
+	// Fullnode loop can start before syncing from DA
 	if !isSequencer {
 		go uevent.MustSubscribe(ctx, m.Pubsub, "applyGossipedBlocksLoop", p2p.EventQueryNewNewGossipedBlock, m.onNewGossipedBlock, m.logger)
+		go m.RetrieveLoop(ctx)
```
mtsitrin, line 149: I think there's issue here.
now `syncUntilTarget` can be called in parallel, which makes no sense.
once from `syncBlockManager` and 2nd from `RetrieveLoop`

srene: true. the retrieve loop should start only when the synblockmanager has finished, otherwise there can be two syncing processes in parallel. i fixed this.

srene: i also moved that sequencers just wait for the da to be synced only to start producing blocks, since blobs can be retrieved even if da is not fully synced. i also renamed the channel to be more clear.

block/manager.go
```diff
@@ -145,23 +144,28 @@ func (m *Manager) Start(ctx context.Context) error {
 	}
 
 	if !isSequencer {
+		// Fullnode loop can start before syncing from DA
 		go uevent.MustSubscribe(ctx, m.Pubsub, "applyGossipedBlocksLoop", p2p.EventQueryNewNewGossipedBlock, m.onNewGossipedBlock, m.logger)
 	}
 
+	// TODO: populate the accumulatedSize on startup
+
+	// Start syncing from DA
```
danwt, line 153: useless comment, if `syncBlockManager` is hard to understand than rename the func instead

danwt, line 153:
```suggestion
```

block/retriever.go
```diff
@@ -38,6 +38,7 @@ func (m *Manager) RetrieveLoop(ctx context.Context) {
 // the actual blocks from the DA.
 func (m *Manager) syncUntilTarget(targetHeight uint64) error {
 	for currH := m.State.Height(); currH < targetHeight; currH = m.State.Height() {
+		m.logger.Info("Syncing from DA", "store height", currH, "target height", targetHeight)
```
danwt, line 41: please put info logs only **after** the thing that happened

so change to debug or moved to end of for loop and put `Synced`

da/celestia/celestia.go
```diff
@@ -133,48 +134,54 @@ func (c *DataAvailabilityLayerClient) Start() (err error) {
...
+				select {
+				case err := <-done:
+					if err != nil {
+						c.logger.Error("Failed to start Celestia DA.", "Error", err)
+					}
 					return err
+				case <-ticker.C:
+					state, err := rpc.Header.SyncState(c.ctx)
+					if err != nil {
+						return err
```
danwt, line 161: want to log this one too?

danwt (changes requested):
looks good just some nits

block/manager.go
danwt: I mean just `// Start syncing from DA`

da/celestia/celestia.go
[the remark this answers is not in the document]
danwt: ok good point

da/celestia/celestia.go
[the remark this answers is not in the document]
danwt: ok fine thanks

block/manager.go
srene: removed

block/retriever.go
srene: done

block/manager.go
```diff
@@ -252,3 +250,14 @@ func getInitialState(store store.Store, genesis *tmtypes.GenesisDoc, logger type
...
+func (m *Manager) startNonSequencerLoop(ctx context.Context) {
+	go uevent.MustSubscribe(ctx, m.Pubsub, "applyGossipedBlocksLoop", p2p.EventQueryNewNewGossipedBlock, m.onNewGossipedBlock, m.logger)
+	go m.RetrieveLoop(ctx)
+	go m.SyncTargetLoop(ctx)
+}
+
+func (m *Manager) startSequencerLoop(ctx context.Context) {
+	go m.ProduceBlockLoop(ctx)
+	go m.SubmitLoop(ctx)
+}
```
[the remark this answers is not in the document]
srene: done

mtsitrin (approved):

da/celestia/celestia.go
```diff
@@ -133,48 +134,55 @@ func (c *DataAvailabilityLayerClient) Start() (err error) {
...
+				case <-ticker.C:
+					state, err := rpc.Header.SyncState(c.ctx)
+					if err != nil {
+						c.logger.Error("Failed to sync Celestia DA.", "Error", err)
+						return err
+					}
+					if !state.Finished() {
+						c.logger.Info("celestia-node still syncing", "height", state.Height, "target", state.ToHeight)
+					}
 				}
```
mtsitrin, line 167: missing return here? in case `state.Finished()` is true

mtsitrin (changes requested):

da/celestia/celestia.go
srene: i think its fine not to return here, since in case there is no error the ticker is just used for logging the syncing progress. in case it finishes it will write to done channel by rpc.Header.SyncWait func and it will exit, or retry in case of error.
however, i realized there is a bug, finalizing before executing the retry function. i fixed it.

mtsitrin: so why checking `state.Finished()` at all? by your logic we could just log "sync in progress" on timer tick.
I guess we wanted a double check on the `SyncWait` signal

block/manager.go
```diff
@@ -150,23 +149,27 @@ func (m *Manager) Start(ctx context.Context) error {
...
 	}
 
+	// TODO: populate the accumulatedSize on startup
+
 	err = m.syncBlockManager()
 	if err != nil {
 		return fmt.Errorf("sync block manager: %w", err)
 	}
 
+	// Sequencer producer loop and full-node retrieve loop is started only when finished syncing from DA
```
omritoptix, line 163: the comment here seems incorrect in regards to the full node

da/celestia/celestia.go
```diff
@@ -629,3 +604,46 @@ func (c *DataAvailabilityLayerClient) getDataAvailabilityHeaders(height uint64)
 
 	return headers.DAH, nil
 }
+
+// Celestia syncing in background
+func (c *DataAvailabilityLayerClient) syncCelestia(rpc *openrpc.Client) {
```
omritoptix, line 609: we're already in celestia so maybe just name the func `sync`?

da/celestia/celestia.go
```diff
@@ -629,3 +604,46 @@ func (c *DataAvailabilityLayerClient) getDataAvailabilityHeaders(height uint64)
...
+		retry.Attempts(0), // try forever
+		retry.Delay(10*time.Second),
+		retry.LastErrorOnly(true),
+		retry.DelayType(retry.FixedDelay),
+	)
+
+	c.logger.Info("Celestia-node is synced.")
+	c.synced <- struct{}{}
+
+	if err != nil {
```
omritoptix, line 646: seems to me like this line will never be reached as we're trying forever.

da/celestia/celestia.go
```diff
@@ -629,3 +604,46 @@ func (c *DataAvailabilityLayerClient) getDataAvailabilityHeaders(height uint64)
...
+					c.logger.Error("Failed to sync Celestia DA.", "Error", err)
+					return err
+				}
+				c.logger.Info("Celestia-node syncing", "height", state.Height, "target", state.ToHeight)
+
+			}
+		}
+	}
+
+	err := retry.Do(sync,
```
omritoptix, line 636: Instead of logging the error in the function I think we can just return a wrapped error and change the retry.Do to something like the following:

```
err := retry.Do(sync,
		retry.Attempts(0), // try forever
		retry.Delay(10*time.Second),
		retry.LastErrorOnly(true),
		retry.DelayType(retry.FixedDelay),
		retry.OnRetry(func(n uint, err error) {
			c.logger.Error("Retry sync due to error", "attempt", n, "error", err)
		}),
	)
```

srene: done

da/celestia/celestia.go
srene: changed

block/manager.go
srene: that was supposed to be removed. fixed

da/celestia/celestia.go
srene: but i guess we need to catch the error of the retry function and do something with it in case it fails, anyway..

omritoptix: ok we can leave it.

omritoptix (approved):

PR #886, opened 2024-05-27
fix(block manage): reducing batch overhead size when reaping Txs from mempool

### LINKED ISSUE #885: Commit size + Block size could be greater than batch size
The issue here is that though we take for the block size the minimum between the batch size and the block size consensus params, in [those lines](https://github.com/dymensionxyz/dymint/blob/main/block/submit.go#L164-L172) we also append the commit to the batch (+ the overhead of the struct of the batch itself) so in some cases the batch won't get populated with any blocks (as it may be that the first block in the batch + commit > max batch size) which would lead to inability to submit batches.

mtsitrin:
# PR Standards

--

---

Close #885 

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/block/submit_test.go
+++ b/block/submit_test.go
@@ -12,83 +12,67 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
-	abci "github.com/tendermint/tendermint/abci/types"
-	cfg "github.com/tendermint/tendermint/config"
-	"github.com/tendermint/tendermint/libs/log"
 	"github.com/tendermint/tendermint/proxy"
-	tmtypes "github.com/tendermint/tendermint/types"
 
 	cosmosed25519 "github.com/cosmos/cosmos-sdk/crypto/keys/ed25519"
 	"github.com/libp2p/go-libp2p/core/crypto"
 
-	"github.com/dymensionxyz/dymint/block"
 	"github.com/dymensionxyz/dymint/config"
-	"github.com/dymensionxyz/dymint/mempool"
-	mempoolv1 "github.com/dymensionxyz/dymint/mempool/v1"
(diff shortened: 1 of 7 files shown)


block/types.go
```diff
@@ -8,6 +8,11 @@ import (
 )
 
 // TODO: move to types package
+
+const (
+	BatchOverhead = 700 // 500 block overhead + 117 for commit + 10 for batch overhead + safety margin
```
omritoptix, line 13: not sure I understand how it was calculated.
also Is this in bytes? can we make it configurable? 
also maybe it can be a percentage/factor of the block size? vs hardcoded bytes?

mtsitrin: refactored it

block/submit_test.go
```diff
@@ -12,17 +12,112 @@ import (
...
+		var blocksize, commitSize int
+		for _, block := range batch.Blocks {
+			blocksize += block.ToProto().Size()
+		}
+		for _, commit := range batch.Commits {
+			commitSize += commit.ToProto().Size()
+		}
+
+		// we assert that the batch size is not larger than the maxBatchSize
+		assert.LessOrEqual(t, batchSize, int(maxBatchSize), tcase.name)
```
danwt, line 111: can you move this up a few lines so that the logging-only computation is clearly demarcated from the actual thing being tested

block/submit_test.go
```diff
@@ -221,3 +316,79 @@ func TestSubmissionByBatchSize(t *testing.T) {
 		}
 	}
 }
+
+func TestFullyPopulatedBlock(t *testing.T) {
```
danwt, line 320: needs description of what it tests

danwt (changes requested):
Looks like it works

I left some readability requests

BTW

Maybe instead of using a threshold like 90%

We can just binary search until we get something that makes maximum use of the space without overfilling. (Using a good start point to note be wasteful)

danwt (approved):
Nice work

omritoptix (approved):

PR #907, opened 2024-06-10
fix(rpc): panic and publish health event only on create batch error

### LINKED ISSUE #902: health endpoint isn't updated if batch fails creation
if e.g failing to create a batch:

`ERRO[43500] handle submission trigger[error create next batch: batch end height must be greater than start height. EndHeight 26643, StartHeight 26644]
`
no health event is emitted and node operator has no visiblity about the node failing to send the batch

zale144:
# PR Standards

--

---

Close #902 

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

## Summary by CodeRabbit

- **New Features**
  - Improved application stability by ensuring blocked processes are properly terminated during shutdown.
  - Enhanced error handling with error logging for better diagnosis and troubleshooting.

- **Performance Improvements**
  - Added health status monitoring for improved system health tracking.


block/submit.go
```diff
@@ -24,8 +24,23 @@
 	maxSizeC := make(chan struct{}, m.Conf.MaxSupportedBatchSkew)
 	go m.AccumulatedDataLoop(ctx, maxSizeC)
 
+	handleSubmitLoopPanic := func() {
+		// recover from panic, emit health status event and return the error
+		var err error
+		if r := recover(); r != nil {
```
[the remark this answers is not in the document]
danwt: @zale144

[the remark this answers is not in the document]
danwt: doesnt work AFAIU

danwt (changes requested):

block/submit.go
```diff
@@ -26,6 +27,17 @@ func (m *Manager) SubmitLoop(ctx context.Context) {
 
 	// defer func to clear the channels to release blocked goroutines on shutdown
 	defer func() {
+		// recover from panic, emit health status event and return the error
```
mtsitrin, line 30: why u handle it in recover. instead of just avoiding the panic here
```
err := m.HandleSubmissionTrigger()
		if err != nil {
			panic(fmt.Errorf("handle submission trigger: %w", err))
		}
```

block/submit.go
```diff
@@ -26,6 +27,17 @@ func (m *Manager) SubmitLoop(ctx context.Context) {
...
+		var err error
+		if r := recover(); r != nil {
+			m.logger.Error(fmt.Errorf("handle submission trigger: %v", r).Error())
+			err, _ = r.(error)
+		}
+		// new context because previous one is canceled
```
mtsitrin, line 36: the defer will return, it won't run the loop again

mtsitrin (changes requested):

block/submit.go
zale144: I guess the idea is that it would be considered an unrecoverable error, so might as well panic. However, we would still like to get an event about it

block/submit.go
zale144: yes

block/submit.go
mtsitrin: @zale144 why we need the event if the rpc server gonna be shut down anyway due to the panic?
don't we need `no panic + event` instead of the existing panic?

mtsitrin (commented):
as discussed, we need to stop production in this case.
maybe the entire blockManager's context can be closed

mtsitrin (approved):

block/manager.go
```diff
@@ -41,6 +43,9 @@ type Manager struct {
 	State    *types.State
 	Executor *Executor
 
+	// Context
+	cancelCtx context.CancelFunc
```
danwt, line 47: This is a complete abuse / anti-pattern of how to write go code and how to use contexts

https://stackoverflow.com/questions/70042646/best-practices-on-go-context-cancelation-functions

You need to do something like instead https://stackoverflow.com/a/70042766

Sorry but this is just wrong golang

danwt (changes requested):

block/manager.go
zale144: Sure there are more idiomatic ways to do this, but I think calling it "complete abuse" of how to write go code is a bit out of line. 
 
Given the time constraints, I don't think these are the most critical issues to deal with right now.

mtsitrin (approved):

Issue #930, opened 2024-06-26
Add support for DA providers using the x/interchain-da module


keruch:
- [x] DA chain client implementation (based on [[1]](https://github.com/dymensionxyz/cosmosclient?tab=readme-ov-file))
- [x] DA chain config
- [x] Interchain-da proto contracts
- [x] Broadcasting transactions (queries, timeouts, attempts, etc)
- [x] SubmitBatch method
- [x] Come up with a commitment structure
- [x] RetrieveBatches method
- [ ] 🔵 Dymension IBC state update <-- design done, impl is in progress
- [ ] Wiring the implementation
- [ ] Support MsgUpdateStateV2
- [ ] ~~(??) Update Dymint to Cosmos SDK v0.50~~

🔵 - in progress

References:
1. [Dymension cosmos client repo](https://github.com/dymensionxyz/cosmosclient?tab=readme-ov-file)
2. [Solution specification](https://www.notion.so/dymension/Dymint-interchain-DA-9e43e85d5c7c4657a127e11ed14c0109)

PR #932, opened 2024-06-27
feat(da): added interchain-da proto contracts

keruch:
Part of #930 

This PR contains modified copies of [interchain-da proto contracts](https://github.com/dymensionxyz/interchain-da/tree/main/proto/dymension/interchain_da). These contracts are needed as part of the implementation of the generalized interchain-da client. We can't directly import these proto files from the interchain-da repo because it uses Cosmos SDK v0.50.6, so the workaround is to copy and adjust all the contracts.

For Author:

- [x]  Wrote unit and integration tests <-- is not needed in this PR
- [x]  All CI checks have passed
- [x]  Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code)

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


proto/types/interchain_da/da.proto
```diff
@@ -0,0 +1,27 @@
...
+
+message Params {
+  // CostPerByte defines the coin cost to store each byte of the blob. This
+  // allows for multidimensional gas pricing on different resources of the
+  // network.
+  cosmos.base.v1beta1.Coin cost_per_byte = 1 [ (gogoproto.nullable) = false ];
+  // MaxBlobSize is the hard cap of how many bytes a blob can be.
+  uint32 max_blob_size = 2;
+  // DisputePeriod is the number of blocks the blob is stored for.
+  uint64 dispute_period = 3;
```
srene, line 20: i would call it something like storage_window or prunning_height, since technically speaking is the time the blob will be stored regardless any dispute

srene (approved):

proto/types/interchain_da/da.proto
danwt: Makes sense @keruch , but you add what you just said as a comment too
thanks

srene (approved):

proto/types/interchain_da/da.proto
danwt: amazing ⭐ thanks

proto/types/cosmos/base/v1beta1/coin.proto
```diff
@@ -0,0 +1,43 @@
...
+
+// IntProto defines a Protobuf wrapper around an Int object.
+message IntProto {
+  string int = 1 [(cosmos_proto.scalar) = "cosmos.Int", (gogoproto.customtype) = "Int", (gogoproto.nullable) = false];
+}
+
+// DecProto defines a Protobuf wrapper around a Dec object.
+message DecProto {
+  string dec = 1 [(cosmos_proto.scalar) = "cosmos.Dec", (gogoproto.customtype) = "Dec", (gogoproto.nullable) = false];
+}
```
danwt, line 43: why the copy paste?

danwt (approved):
Cool stuff
I'm not super familiar with the proto setup though, what's the need for the redefinitions of existing cosmos stuff?

proto/types/cosmos/base/v1beta1/coin.proto
danwt: okok
guess we should fix that at somepoint

Issue #963, opened 2024-07-18
In   AccumulatedDataLoop miscounting -  do not assume batches are all BlockBatchMaxSizeBytes  size


danwt:
Even after

- https://github.com/dymensionxyz/dymint/pull/962
- https://github.com/dymensionxyz/dymint/pull/960

which fix

- https://github.com/dymensionxyz/dymint/issues/959
- https://github.com/dymensionxyz/dymint/issues/954

there is still a problem:

The AccumulatedDataLoop basically assumes that the size of the batch being submitted is equal to BlockBatchMaxSizeBytes but actually it will be less. That means every time the submission happens it is actually falling more and more behind the actual produced blocks.

Issue #968, opened 2024-07-18
Accumulated data loop does not account for blocks produced by time


danwt:
After the merging of this PR https://github.com/dymensionxyz/dymint/pull/960 

the loop is not aware that sometimes batches might be submitted by time

PR #969, opened 2024-07-18
fix(submission): fix counting and time

danwt:
# PR Standards

--

---

Close https://github.com/dymensionxyz/dymint/issues/963, https://github.com/dymensionxyz/dymint/issues/968, https://github.com/dymensionxyz/dymint/issues/973

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/block/manager.go
+++ b/block/manager.go
@@ -158,7 +158,7 @@ func (m *Manager) Start(ctx context.Context) error {
 		// Sequencer must wait till DA is synced to start submitting blobs
 		<-m.DAClient.Synced()
 
-		bytesProducedC := make(chan int64, 1)
+		bytesProducedC := make(chan int, 1)
 		eg.Go(func() error {
 			bytesProducedC <- m.GetUnsubmittedBytes()
 			return m.SubmitLoop(ctx, bytesProducedC)
--- a/block/manager.go
+++ b/block/manager.go
@@ -158,7 +158,7 @@ func (m *Manager) Start(ctx context.Context) error {
 		// Sequencer must wait till DA is synced to start submitting blobs
 		<-m.DAClient.Synced()
 
-		bytesProducedC := make(chan int64)
+		bytesProducedC := make(chan int64, 1)
 		eg.Go(func() error {
 			bytesProducedC <- m.GetUnsubmittedBytes()
 			return m.SubmitLoop(ctx, bytesProducedC)
--- a/block/manager.go
+++ b/block/manager.go
@@ -152,18 +152,19 @@ func (m *Manager) Start(ctx context.Context) error {
 		return fmt.Errorf("sync block manager: %w", err)
 	}
 
-	productionC := make(chan struct{})
-
 	eg, ctx := errgroup.WithContext(ctx)
 
 	if isSequencer {
 		// Sequencer must wait till DA is synced to start submitting blobs
 		<-m.DAClient.Synced()
+
+		bytesProducedC := make(chan int64)
 		eg.Go(func() error {
-			return m.SubmitLoop(ctx, productionC)
+			bytesProducedC <- m.GetUnsubmittedBytes()
+			return m.SubmitLoop(ctx, bytesProducedC)
 		})
 		eg.Go(func() error {
-			return m.ProduceBlockLoop(ctx, productionC)
+			return m.ProduceBlockLoop(ctx, bytesProducedC)
 		})
 	} else {
 		eg.Go(func() error {
--- a/block/manager.go
+++ b/block/manager.go
@@ -223,19 +223,3 @@ func (m *Manager) syncBlockManager() error {
 	m.logger.Info("Synced.", "current height", m.State.Height(), "last submitted height", m.LastSubmittedHeight.Load())
 	return nil
 }
-
-func (m *Manager) MustLoadBlock(h uint64) *types.Block {
-	ret, err := m.Store.LoadBlock(h)
-	if err != nil {
-		panic(fmt.Errorf("store load block: height: %d: %w", h, err))
-	}
-	return ret
-}
-
-func (m *Manager) MustLoadCommit(h uint64) *types.Commit {
-	ret, err := m.Store.LoadCommit(h)
-	if err != nil {
-		panic(fmt.Errorf("store load commit: height: %d: %w", h, err))
-	}
-	return ret
-}
--- a/block/manager.go
+++ b/block/manager.go
@@ -159,8 +159,8 @@ func (m *Manager) Start(ctx context.Context) error {
 		<-m.DAClient.Synced()
 
 		bytesProducedC := make(chan int, 1)
+		bytesProducedC <- m.GetUnsubmittedBytes()
 		eg.Go(func() error {
-			bytesProducedC <- m.GetUnsubmittedBytes()
 			return m.SubmitLoop(ctx, bytesProducedC)
 		})
 		eg.Go(func() error {
--- a/block/manager.go
+++ b/block/manager.go
@@ -157,9 +157,11 @@ func (m *Manager) Start(ctx context.Context) error {
 	if isSequencer {
 		// Sequencer must wait till DA is synced to start submitting blobs
 		<-m.DAClient.Synced()
-
-		bytesProducedC := make(chan int, 1)
-		bytesProducedC <- m.GetUnsubmittedBytes()
+		nBytes := m.GetUnsubmittedBytes()
+		bytesProducedC := make(chan int)
+		go func() {
+			bytesProducedC <- nBytes
+		}()
 		eg.Go(func() error {
 			return m.SubmitLoop(ctx, bytesProducedC)
 		})
--- a/block/manager.go
+++ b/block/manager.go
@@ -51,11 +51,6 @@ type Manager struct {
 	SLClient  settlement.ClientI
 
 	/*
-		Production
-	*/
-	producedSizeC chan uint64 // for the producer to report the size of the block+commit it produced
-
-	/*
 		Submission
 	*/
 	// The last height which was submitted to both sublayers, that we know of. When we produce new batches, we will
@@ -114,7 +109,6 @@ func NewManager(
 		SLClient:         settlementClient,
 		Retriever:        dalc.(da.BatchRetriever),
 		targetSyncHeight: diodes.NewOneToOne(1, nil),
-		producedSizeC:    make(chan uint64),
 		logger:           logger,
 		blockCache:       make(map[uint64]CachedBlock),
 	}
@@ -158,16 +152,18 @@ func (m *Manager) Start(ctx context.Context) error {
 		return fmt.Errorf("sync block manager: %w", err)
 	}
 
+	productionC := make(chan struct{})
+
 	eg, ctx := errgroup.WithContext(ctx)
 
 	if isSequencer {
 		// Sequencer must wait till DA is synced to start submitting blobs
 		<-m.DAClient.Synced()
 		eg.Go(func() error {
-			return m.SubmitLoop(ctx)
+			return m.SubmitLoop(ctx, productionC)
 		})
 		eg.Go(func() error {
-			return m.ProduceBlockLoop(ctx)
+			return m.ProduceBlockLoop(ctx, productionC)
 		})
 	} else {
 		eg.Go(func() error {
--- a/block/produce.go
+++ b/block/produce.go
@@ -21,7 +21,9 @@ import (
 )
 
 // ProduceBlockLoop is calling publishBlock in a loop as long as we're synced.
-func (m *Manager) ProduceBlockLoop(ctx context.Context) error {
+// A signal will be sent to the wait channel for each block produced
+// In this way it's possible to pause block production by not consuming the channel
+func (m *Manager) ProduceBlockLoop(ctx context.Context, wait chan struct{}) error {
 	m.logger.Info("Started block producer loop.")
 
 	ticker := time.NewTicker(m.Conf.BlockTime)
@@ -42,7 +44,7 @@ func (m *Manager) ProduceBlockLoop(ctx context.Context) error {
 			produceEmptyBlock := firstBlock || 0 == m.Conf.MaxIdleTime || nextEmptyBlock.Before(time.Now())
 			firstBlock = false
 
-			block, commit, err := m.ProduceAndGossipBlock(ctx, produceEmptyBlock)
+			block, err := m.ProduceAndGossipBlock(ctx, produceEmptyBlock)
 			if errors.Is(err, context.Canceled) {
 				m.logger.Error("Produce and gossip: context canceled.", "error", err)
 				return nil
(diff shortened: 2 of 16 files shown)


block/produce.go
```diff
@@ -36,51 +38,56 @@ func (m *Manager) ProduceBlockLoop(ctx context.Context) (err error) {
...
+				// TODO: optimize to only do this if IBC transactions are present (https://github.com/dymensionxyz/dymint/issues/709)
 				nextEmptyBlock = time.Now().Add(m.Conf.MaxProofTime)
 			} else {
 				m.logger.Info("Produced empty block.")
 			}
 
-			// Send the size to the accumulated size channel
-			// This will block in case the submitter is too slow and it's buffer is full
-			size := uint64(block.ToProto().Size()) + uint64(commit.ToProto().Size())
+			pause := len(bytesProducedC) == cap(bytesProducedC) // here we assume cap is at least 1
```
zale144, line 73: I think something like this would be more concise and easier to understand:

```
select {
	case <-ctx.Done():
		return nil
	case bytesProducedC <- block.SizeBytes() + commit.SizeBytes():
	default:
		evt := &events.DataHealthStatus{Error: fmt.Errorf("bytes produced channel is full: %w", gerrc.ErrResourceExhausted)}
		uevent.MustPublish(ctx, m.Pubsub, evt, events.HealthStatusList)
		m.logger.Error("Enough bytes to build a batch have been accumulated, but too many batches are pending submission. Pausing block production until a signal is consumed.")

		bytesProducedC <- block.SizeBytes() + commit.SizeBytes()
		evt = &events.DataHealthStatus{Error: nil}
		uevent.MustPublish(ctx, m.Pubsub, evt, events.HealthStatusList)
		m.logger.Info("Resumed block production.")
}
```

block/submit.go
```diff
@@ -2,200 +2,170 @@ package block
...
-// Finally, it submits the next batch of blocks and updates the sync target to the height of the last block in the submitted batch.
-func (m *Manager) HandleSubmissionTrigger() error {
-	// Load current sync target and height to determine if new blocks are available for submission.
-
-	startHeight := m.NextHeightToSubmit()
-	endHeightInclusive := m.State.Height()
-
-	if endHeightInclusive < startHeight {
-		return nil // No new blocks have been produced
+func (m *Manager) CreateAndSubmitBatchGetSizeEstimate() (int64, error) {
```
zale144: did this really have to be one method?

types/serialization.go
```diff
@@ -195,8 +197,15 @@ func (b *Block) FromProto(other *pb.Block) error {
...
+	end := other.EndHeight
+	if 0 < n && start != other.Blocks[0].Header.GetHeight() {
+		return fmt.Errorf("start height does not match first block height: %w", gerrc.ErrInvalidArgument)
+	}
+	if 0 < n && end != other.Blocks[n-1].Header.GetHeight() {
+		return fmt.Errorf("end height does not match last block height: %w", gerrc.ErrInvalidArgument)
+	}
 	b.Blocks = protoToBlocks(other.Blocks)
 	b.Commits = protoToCommits(other.Commits)
 	return nil
```
zale144, line 211:
```suggestion
	b.Blocks = protoToBlocks(other.Blocks)
	b.Commits = protoToCommits(other.Commits)
	
	n := len(other.Blocks)
	if n == 0 {
		return
	}

	if other.StartHeight != other.Blocks[0].Header.GetHeight() {
		return fmt.Errorf("start height does not match first block height: %w", gerrc.ErrInvalidArgument)
	}
	if  other.EndHeight != other.Blocks[n-1].Header.GetHeight() {
		return fmt.Errorf("end height does not match last block height: %w", gerrc.ErrInvalidArgument)
	}
```

block/production_test.go
```diff
@@ -57,13 +58,11 @@ func TestCreateEmptyBlocksEnableDisable(t *testing.T) {
...
-
-	buf1 := make(chan struct{}, 100) // dummy to avoid unhealthy event
-	buf2 := make(chan struct{}, 100) // dummy to avoid unhealthy event
-	go manager.AccumulatedDataLoop(mCtx, buf1)
-	go managerWithEmptyBlocks.AccumulatedDataLoop(mCtx, buf2)
+	bytesProduced1 := make(chan int)
+	bytesProduced2 := make(chan int)
+	go manager.ProduceBlockLoop(mCtx, bytesProduced1)
+	go managerWithEmptyBlocks.ProduceBlockLoop(mCtx, bytesProduced2)
+	uchannel.DrainForever(bytesProduced1, bytesProduced2)
```
zale144, line 65: why is draining needed here?

block/submit.go
danwt: It's more testable like this

block/produce.go
danwt: That looks wrong because you are not waiting on context on the second `bytesProducedC <- block.SizeBytes() + commit.SizeBytes()` line

types/serialization.go
danwt: 🤷

block/production_test.go
danwt: to stop the producer being blocked

block/manager.go
```diff
@@ -163,11 +157,14 @@ func (m *Manager) Start(ctx context.Context) error {
 	if isSequencer {
 		// Sequencer must wait till DA is synced to start submitting blobs
 		<-m.DAClient.Synced()
+
+		bytesProducedC := make(chan int, 1)
```
mtsitrin, line 161: why size 1? it can lead to health toggling in some race conditions

block/submit.go
```diff
@@ -2,200 +2,210 @@ package block
...
-			default:
-				return
+			if maxBatchSkew*maxBatchBytes < pendingBytes.Load() {
+				// too much stuff is pending submission
+				// we block here until we get a progress nudge from the submitter thread
+				select {
+				case <-ctx.Done():
+					return ctx.Err()
+				case <-counter.C:
+				case <-ticker.C:
```
mtsitrin, line 61: if u stopped on `too much stuff pending`, why u unblock in case of `ticker.C`?

block/submit.go
```diff
@@ -2,200 +2,210 @@ package block
...
-				select {
-				case <-ctx.Done():
-					return
-				case toSubmit <- struct{}{}:
+				return ctx.Err()
+			case <-submitter.C:
+			}
+			pending := pendingBytes.Load()
+			// while there are accumulated blocks, create and submit batches!!
+			for ctx.Err() == nil && (0 < pending && (maxBatchTime < time.Since(timeLastSubmission)) || maxBatchBytes < pending) {
```
mtsitrin, line 90: pls break down condition to more readable expression

mtsitrin: u can also `continue` on pending = 0

block/submit.go
```diff
@@ -2,200 +2,210 @@ package block
...
+	})
 
-	startHeight := m.NextHeightToSubmit()
-	endHeightInclusive := m.State.Height()
+	return eg.Wait()
+}
 
-	if endHeightInclusive < startHeight {
-		return nil // No new blocks have been produced
+func (m *Manager) CreateAndSubmitBatchGetSizeEstimate(maxSize uint64) (uint64, error) {
```
mtsitrin, line 105: `GetSizeEstimate` in the func name is redundent.
u can name the `(uint64, error)` if for explicitily if u wish
godoc

block/submit.go
```diff
@@ -2,200 +2,210 @@ package block
...
-	m.logger.Info("Submitted batch to SL.", "start height", nextBatch.StartHeight, "end height", actualEndHeight)
-
-	types.RollappHubHeightGauge.Set(float64(actualEndHeight))
-	m.LastSubmittedHeight.Store(actualEndHeight)
-	return nil
+	return b, nil
 }
 
-func (m *Manager) CreateNextBatchToSubmit(startHeight uint64, endHeightInclusive uint64) (*types.Batch, error) {
+func CreateBatch(store store.Store, maxBatchSize uint64, startHeight uint64, endHeightInclusive uint64) (*types.Batch, error) {
```
mtsitrin, line 130: why it;'s not `func (m *Manager)` method?

mtsitrin (changes requested):

types/batch.go
```diff
@@ -7,8 +7,45 @@ const (
...
+
+// NumBlocks is the number of blocks in the batch
+func (b Batch) NumBlocks() uint64 {
+	return uint64(len(b.Blocks))
+}
+
+// SizeBytesEstimate returns the sum of the size of bytes of the blocks and commits
+// The actual size of the batch may be different due to additional metadata and protobuf
+// optimizations.
+func (b Batch) SizeBytesEstimate() int {
```
omritoptix, line 38: I think the name is a bit confusing as it's the size of the inner blocks + commits, which is important to be accurate and not estimate to avoid a every growing tail. 
I'd suggest something like `SizeBlockAndCommitBytes` or `SizeInternalBytes` or somethign which doesn't say `estimate`.

block/submit.go
```diff
@@ -2,200 +2,210 @@ package block
...
+				if err != nil {
+					return fmt.Errorf("create and submit batch: %w", err)
 				}
-
-				evt = &events.DataHealthStatus{Error: nil}
-				uevent.MustPublish(ctx, m.Pubsub, evt, events.HealthStatusList)
-
-				m.logger.Info("Resumed block production.")
+				timeLastSubmission = time.Now()
+				pending = pendingBytes.Add(^(nConsumed - 1)) // subtract
```
omritoptix, line 96: shouldn't it be `pending = pendingBytes.Add(^uint64(nConsumed - 1) + 1) // subtract` ?

block/submit.go
```diff
@@ -2,200 +2,210 @@ package block
...
+		ticker := time.NewTicker(maxBatchTime)
+		defer ticker.Stop()
 		for {
-			select {
-			case <-m.producedSizeC:
-			case <-maxSizeC:
-			default:
-				return
+			if maxBatchSkew*maxBatchBytes < pendingBytes.Load() {
+				// too much stuff is pending submission
```
omritoptix, line 55: not sure I see where do we emit an health event in case of batch skew too big?

omritoptix (changes requested):

block/submit.go
danwt: on producer thread, which makes more sense IMO
but super bullish on it

block/submit.go
```diff
@@ -2,200 +2,210 @@ package block
...
 	if err != nil {
-		return fmt.Errorf("create next batch to submit: %w", err)
+		return nil, fmt.Errorf("create batch: %w", err)
 	}
 
-	resultSubmitToDA := m.DAClient.SubmitBatch(nextBatch)
-	m.logger.Info("Submitted batch to DA", "start height", nextBatch.StartHeight, "end height", nextBatch.EndHeight)
-	if resultSubmitToDA.Code != da.StatusSuccess {
-		return fmt.Errorf("submit next batch to da: %s", resultSubmitToDA.Message)
+	if 0 == b.NumBlocks() {
```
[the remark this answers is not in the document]
danwt: Hmm ok nice, I think this is achieved if i simply remove the `if`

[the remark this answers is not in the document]
danwt: gonna tackle later https://github.com/dymensionxyz/dymint/pull/969

block/submit.go
```diff
@@ -2,200 +2,210 @@ package block
...
-				case <-ctx.Done():
-					return
-				case toSubmit <- struct{}{}:
+				return ctx.Err()
+			case <-submitter.C:
+			}
+			pending := pendingBytes.Load()
+			// while there are accumulated blocks, create and submit batches!!
+			for ctx.Err() == nil && (0 < pending && (maxBatchTime < time.Since(timeLastSubmission)) || maxBatchBytes < pending) {
+				nConsumed, err := createAndSubmitBatchGetSizeEstimate(min(pending, maxBatchBytes))
```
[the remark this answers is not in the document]
danwt: Re ctx.Err() - needed to break loop if context cancelled or something

re submit empty: ok makes sense but will do in another pr!

https://github.com/dymensionxyz/dymint/issues/980

block/submit.go
danwt: >u can also continue on pending = 0
I dont' get you

>pls break down condition to more readable expression
Ok I have made a change

block/submit.go
danwt: I followed omris suggestion instead, but good point

block/submit.go
```diff
@@ -2,200 +2,210 @@ package block
...
+	createAndSubmitBatchGetSizeEstimate func(maxSize uint64) (uint64, error),
+) error {
+	eg, ctx := errgroup.WithContext(ctx)
+
+	pendingBytes := atomic.Uint64{}
+	counter := uchannel.NewNudger()   // used to avoid busy waiting (using cpu) on counter thread
+	submitter := uchannel.NewNudger() // used to avoid busy waiting (using cpu) on submitter thread
+
+	eg.Go(func() error {
+		// 'counter': we need one thread to continuously consume the bytes produced channel, and to monitor timer
```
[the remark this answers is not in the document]
danwt: changed

block/submit.go
danwt: done

utils/channel/funcs.go
```diff
@@ -0,0 +1,29 @@
+package channel
+
+func DrainForever[T any](chs ...<-chan T) {
```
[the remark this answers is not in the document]
danwt: done

utils/event/funcs.go
```diff
@@ -22,7 +24,7 @@ func MustSubscribe(
 	logger types.Logger,
 ) {
 	subscription, err := pubsubServer.SubscribeUnbuffered(ctx, clientID, eventQuery)
-	if err != nil {
+	if err != nil && !errors.Is(err, context.Canceled) && !errors.Is(err, context.DeadlineExceeded) {
```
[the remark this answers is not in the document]
danwt: good point, done

block/submit.go
```diff
@@ -2,200 +2,210 @@ package block
...
+// It will back pressure (pause) block production if it falls too far behind.
+func (m *Manager) SubmitLoop(ctx context.Context,
+	bytesProduced chan int,
+) (err error) {
+	return SubmitLoopInner(ctx,
+		bytesProduced,
+		m.Conf.MaxBatchSkew,
+		m.Conf.BatchSubmitMaxTime,
+		m.Conf.BatchMaxSizeBytes,
+		m.CreateAndSubmitBatchGetSizeEstimate,
```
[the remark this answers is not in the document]
danwt: This unit of logic is more testable as I have written it with injections
We should stop throwing everything on manager object and requiring so much machinery to test things

block/submit.go
```diff
@@ -2,200 +2,210 @@ package block
...
+	maxBatchBytes uint64, // max size of serialised batch in bytes
+	createAndSubmitBatchGetSizeEstimate func(maxSize uint64) (uint64, error),
+) error {
+	eg, ctx := errgroup.WithContext(ctx)
+
+	pendingBytes := atomic.Uint64{}
+	counter := uchannel.NewNudger()   // used to avoid busy waiting (using cpu) on counter thread
+	submitter := uchannel.NewNudger() // used to avoid busy waiting (using cpu) on submitter thread
+
+	eg.Go(func() error {
```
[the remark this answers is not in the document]
danwt: >encapsulte in seperate funcs (counter, submitter)

IMO in this case it would not be more readable, as it is written it's quite easy to see all the logic one screen page, and there is no advantage in testing from splitting it out. Also, they both use the three variables pendingBytes, trigger, submitter, so they'd need to be passed as args

> IMO SubmitLoopInner can be merged back with SubmitLoop

In this case I disagree.  SubmitLoopInner  is very testable
It's a pretty common pattern to have an inner impl with testable DI and a wrapper (SubmitLoop) to use defaults

block/submit.go
danwt: Great question actually, I've added a comment

https://github.com/dymensionxyz/dymint/blob/e984d293e3af1895f2e8de44e675e3798b2b30b4/block/submit.go#L60-L65

block/manager.go
danwt: Hmm, I will change to make it cleaner, but what specific problem do you see? Health toggling can occur regardless

omritoptix (approved):

block/manager.go
mtsitrin: on some cases, u might produce 2 blocks while the submit loop as a consumer won't read it fast enough for some reason.

mtsitrin (approved):

PR #971, opened 2024-07-18
fix(rpc): Fix status `CatchingUp` field updating

### LINKED ISSUE #935: `/status` endpoints `sync_info.catching_up` field always shows false
`/status` endpoints `sync_info.catching_up` field always shows false, even when a full node is behind

zale144:
# PR Standards

--

---

Close #935

Basically, the idea is to have the CatchingUp value be derived as such:

`CatchingUp = targetHeight > latestHeight`

My guess is that there might need to be some tolerance there, in case a new settlement batch is submitted, and at the same moment the `/status` is queried.

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


rpc/client/client.go
```diff
@@ -751,13 +748,13 @@ func (c *Client) Status(ctx context.Context) (*ctypes.ResultStatus, error) {
 			LatestAppHash:     latestAppHash[:],
 			LatestBlockHeight: int64(latestHeight),
 			LatestBlockTime:   time.Unix(0, int64(latestBlockTimeNano)),
+			CatchingUp:        c.node.BlockManager.TargetHeight.Load() > latestHeight,
```
mtsitrin, line 751: u need to set `TargetHeight` also in  `syncBlockManager` , currnetly u made it event based only

block/manager.go
```diff
@@ -64,6 +64,7 @@ type Manager struct {
 	// It is ALSO used by the producer, because the producer needs to check if it can prune blocks and it wont'
 	// prune anything that might be submitted in the future. Therefore, it must be atomic.
 	LastSubmittedHeight atomic.Uint64
+	TargetHeight        atomic.Uint64
```
mtsitrin, line 67: I think it's redundent. both `LastSubmittedHeight` and `TargetHeight` are actually `LatestHeightOnSL`

mtsitrin (changes requested):

block/gossip.go
```diff
@@ -22,6 +23,10 @@ func (m *Manager) onNewGossipedBlock(event pubsub.Message) {
 		return
 	}
 
+	if block.Header.Height > m.TargetHeight.Load() {
+		m.TargetHeight.Store(block.Header.Height)
+	}
```
danwt, line 28: this is not concurrency safe you need to use compareAndSwap
and please use stdlib max

block/manager.go
```diff
@@ -55,9 +55,10 @@ type Manager struct {
 	*/
 	// The last height which was submitted to both sublayers, that we know of. When we produce new batches, we will
 	// start at this height + 1.
-	// It is ALSO used by the producer, because the producer needs to check if it can prune blocks and it wont'
+	// It is ALSO used by the producer, because the producer needs to check if it can prune blocks and it won't
 	// prune anything that might be submitted in the future. Therefore, it must be atomic.
 	LastSubmittedHeight atomic.Uint64
+	TargetHeight        atomic.Uint64
```
danwt, line 61: It's under the submission section but it's used in the retriever loop

Needs a docstring too, I don't understand how it relates (or doesn't) to targetSyncHeight

danwt (changes requested):

rpc/client/client.go
danwt: this field needs a docstring as well IMO
the other ones are obvious and fine

danwt (changes requested):
I'm still not sure how this relates to the targetSyncHeight field and why we can just have one field

rpc/client/client.go
zale144: fix pushed

rpc/client/client.go
```diff
@@ -750,13 +747,14 @@ func (c *Client) Status(ctx context.Context) (*ctypes.ResultStatus, error) {
 			LatestAppHash:     latestAppHash[:],
 			LatestBlockHeight: int64(latestHeight),
 			LatestBlockTime:   time.Unix(0, int64(latestBlockTimeNano)),
+			// CatchingUp is true if the node is not at the latest height received from p2p or da.
```
danwt, line 750: shuld be on the field on the struct, not where the value is set!

zale144: should I fork Tendermint then?

danwt: ah sorry didn't realise it was tendermint
all good

block/manager.go
```diff
@@ -225,3 +228,12 @@ func (m *Manager) syncBlockManager() error {
 	m.logger.Info("Synced.", "current height", m.State.Height(), "last submitted height", m.LastSubmittedHeight.Load())
 	return nil
 }
+
+func (m *Manager) updateTargetHeight(h uint64) {
+	for {
+		currentHeight := m.TargetHeight.Load()
+		if m.TargetHeight.CompareAndSwap(currentHeight, max(currentHeight, h)) {
```
mtsitrin, line 235: if `max(currentHeight, h)` is `currentHeight`
u'll call `m.TargetHeight.CompareAndSwap(currentHeight,currentHeight)`
won't it return `false` and be stuck forever?

mtsitrin (approved):

PR #974, opened 2024-07-19
fix(p2p): set gossipsub buffersize to avoid missed blocks when applying big number of blocks in a row (hotfix)

### LINKED ISSUE #972: p2p gossiped blocks lost when applying batches from DA
this happens when full-nodes are out of sync, and are receiving batches from DA and blocks gossiped from p2p at the same time, when the received block rate is high and DA batches big.

When applying batches from DA there is a mutex here https://github.com/dymensionxyz/dymint/blob/41d6053509d7c1297d86bc52a66fdefb9c565b1a/block/retriever.go#L116

This causes the new gossiped block event not consumed (we use unbuffered pubsub channel so its blocking)
https://github.com/dymensionxyz/dymint/blob/41d6053509d7c1297d86bc52a66fdefb9c565b1a/block/gossip.go#L17

That at the same time is blocking the func that is consuming received blocks and sending events
https://github.com/dymensionxyz/dymint/blob/41d6053509d7c1297d86bc52a66fdefb9c565b1a/p2p/client.go#L402

And it makes the channel used at  gossipsub level to consume received blocks overflow 
https://github.com/dymensionxyz/go-libp2p-pubsub/blob/3ecd83c19ea24b01677ad5a280b2b2c9a6b4ce87/topic.go#L164

So i guess the easiest solution is to define a bigger buffer size for the gossipsub channel here:
https://github.com/dymensionxyz/dymint/blob/41d6053509d7c1297d86bc52a66fdefb9c565b1a/p2p/gossip.go#L54

like this: 
	subscription, err := topic.Subscribe(pubsub.WithBufferSize(5000))

or to use buffered subscription in the pubsub server again for the received gossip block events that is big enough.

srene:
# PR Standards

--

---

Close #972 

There is a channel in gossipsub implementation, used to consume packets received, that its default size is 32. 
If the occupancy goes over the limit, packets are not added and dropped. 
By setting a bigger buffer size for the channel  we limit the problem, which was happening in cases where out of sync full-nodes where trying to apply long list of blocks, while receiving them from p2p at fast rate.

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/p2p/client.go
+++ b/p2p/client.go
@@ -41,7 +41,7 @@ const (
 	// blockTopicSuffix is added after namespace to create pubsub topic for block gossiping.
 	blockTopicSuffix = "-block"
 
-	// buffer size used when in gossipSub to consume receive packets. packets are dropped in case buffer overflows. at a block rate of 100ms (the min accepted) it can buffer up to  5 min of blocks.
+	// buffer size used when in gossipSub to consume receive packets. packets are dropped in case buffer overflows. at a block rate of 200ms (the min accepted) it can buffer up to  10 min of blocks.
 	pubsubBufferSize = 3000
 )
 
--- a/p2p/client.go
+++ b/p2p/client.go
@@ -40,6 +40,9 @@ const (
 
 	// blockTopicSuffix is added after namespace to create pubsub topic for block gossiping.
 	blockTopicSuffix = "-block"
+
+	// buffer size used when consuming packets received from gossipSub. at a min 100ms blockrate it can buffer up to  5 min block
+	pubsubBufferSize = 3000
 )
 
 // Client is a P2P client, implemented with libp2p.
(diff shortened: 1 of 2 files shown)


artemijspavlovs (approved):

p2p/client.go
```diff
@@ -40,6 +40,9 @@ const (
 
 	// blockTopicSuffix is added after namespace to create pubsub topic for block gossiping.
 	blockTopicSuffix = "-block"
+
+	// buffer size used when in gossipSub to consume receive packets. packets are dropped in case buffer overflows. at a block rate of 100ms (the min accepted) it can buffer up to  5 min of blocks.
```
omritoptix, line 44:
```suggestion
	// buffer size used when in gossipSub to consume receive packets. packets are dropped in case buffer overflows. at a block rate of 200ms (the min accepted) it can buffer up to  10 min of blocks.
```

omritoptix (commented):
If we want only 5 mins, given assumption of 200ms, so we can reduce to 1500, or alternatively keep it 3000 and change it to 10 mins.

srene:
closing since its duplicated and not needed

PR #975, opened 2024-07-19
fix(p2p): set gossipsub buffersize to avoid missed blocks

srene:
# PR Standards

--

---

Close https://github.com/dymensionxyz/dymint/issues/972

There is a channel in gossipsub implementation, used to consume packets received, that its default size is 32.
If the occupancy goes over the limit, packets are not added and dropped.
By setting a bigger buffer size for the channel we limit the problem, which was happening in cases where out of sync full-nodes where trying to apply long list of blocks, while receiving them from p2p at fast rate.

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


artemijspavlovs (approved):

omritoptix (approved):

p2p/gossip.go
```diff
@@ -50,8 +53,7 @@ func NewGossiper(host host.Host, ps *pubsub.PubSub, topicStr string, msgHandler
 	if err != nil {
 		return nil, err
 	}
-
-	subscription, err := topic.Subscribe()
+	subscription, err := topic.Subscribe(pubsub.WithBufferSize(max(pubsub.GossipSubHistoryGossip, pubsubBufferSize)))
```
mtsitrin, line 56: not sure why we have `GossipSubCacheSize: 50` default, if take a max with `pubsubBufferSize = 3000`.
Why not setting the default to be 3000 and just use the `GossipSubCacheSize` conf param?

mtsitrin: if u wish to enforce it, u can check it in 
`(c P2PConfig) Validate()`, but I think it's unneeded

srene: i dont think is a good idea to set it to the same value, since GossipSubCacheSize has another (network) implications, such a periodic overhead that nodes need to send to advertise cached packets, while pubsubBufferSize its just affecting the local buffer size for non-consumed packets. Ideally, we want to keep GossipSubCacheSize smaller than pubsubBufferSize.

mtsitrin (approved):

PR #979, opened 2024-07-19
feat(rpc): Add sync info metrics

### LINKED ISSUE #978: Add debug syncing commands for dymint
The following cli command should reveal:

- [ ] height of last block received from p2p
- [ ] height of last block received from da
- [ ] last block applied source (da or gossip)
- [ ] number of accumulated blocks pending to be applied (cache occupancy)
- [ ] height of lowest block pending in cache
- [ ] height of highest received block

zale144:
# PR Standards

--

---

Close #978 

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


block/retriever.go
```diff
@@ -116,10 +114,9 @@ func (m *Manager) ProcessNextDABatch(daMetaData *da.DASubmitMetaData) error {
 		return batchResp.Error
 	}
 
-	m.logger.Debug("retrieved batches", "n", len(batchResp.Batches), "daHeight", daMetaData.Height)
+	m.LastReceivedDAHeight.Store(daMetaData.Height)
 
-	m.retrieverMu.Lock()
```
srene, line 121: I think removing the retrieverMu usage here could be an issue, since i think it is used for two reasons. protecting against concurrent access to the blockCache, and concurrent block application. if we remove it here it could be possible to applyblocks here and at attemptApplyCachedBlocks at the same time. so i would keep it or i would define another mutex for block application

block/manager.go
```diff
@@ -236,3 +242,43 @@ func (m *Manager) MustLoadCommit(h uint64) *types.Commit {
 	}
 	return ret
 }
+
+func (m *Manager) AddBlockToCache(h uint64, b *types.Block, c *types.Commit) {
+	m.retrieverMu.Lock()
```
omritoptix, line 247: I think better to change the lock to `blockCacheMu` as it's only being used for block cache.
I think one of the pr does it already.
Also I may move all of its to a seperate file called blockCache.

danwt (changes requested):
Nice idea to factor out the cache, but can we stop adding everything directly to manager. And I don't think protecting the cache was the only job of the retrieverMu

Please define a new Cache type which has a mutex and a map

Also +1 Omri's prometheus suggestion

block/retriever.go
omritoptix: agree with sergi. we need a mutex for block application.

block/retriever.go
```diff
@@ -98,12 +98,10 @@ func (m *Manager) applyLocalBlock(height uint64) error {
 		return fmt.Errorf("validate block from local store: height: %d: %w", height, err)
 	}
 
-	m.retrieverMu.Lock()
```
omritoptix, line 101: same here. we need a mutex for block application to make sure no 2 blocks can be applied at the same time.

block/block.go
```diff
@@ -114,19 +116,16 @@ func (m *Manager) isHeightAlreadyApplied(blockHeight uint64) (bool, error) {
 }
 
 func (m *Manager) attemptApplyCachedBlocks() error {
-	m.retrieverMu.Lock()
```
omritoptix, line 117: by removing this, we basically can't know for sure that the `expectedHeight` from the time we get it, till the time we try to apply it, is correct (as another method could apply a different block within this time).
This in turn would fail the validation, which currently raises an error. 

so we either need to gracefully handle this, or assure the `expectedHeight` can't be changed between getting it to try to apply the block.

rpc/client/client.go
```diff
@@ -769,6 +770,28 @@ func (c *Client) Status(ctx context.Context) (*ctypes.ResultStatus, error) {
 	return result, nil
 }
 
+// SyncInfo returns detailed information about current sync status.
+func (c *Client) SyncInfo() *ResultSyncInfo {
```
omritoptix, line 774: thinking, as to my previous comments if all this info just better off written to a gauge

block/block_cache.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func (m *Cache) DeleteBlockFromCache(h uint64) {
+	m.Lock()
+	defer m.Unlock()
+	delete(m.cache, h)
+
+	types.NumberOfAccumulatedBlocks.Set(float64(len(m.cache)))
+	if h == m.lowestCachedBlockHeight {
+		m.lowestCachedBlockHeight = h + 1
```
omritoptix, line 39: you're assuming blocks in cache are by order, but you can have a cache with heights e.g [5,8,12]

block/block_cache.go
```diff
@@ -0,0 +1,64 @@
...
+		types.LowestPendingBlockHeight.Set(float64(m.lowestCachedBlockHeight))
+	}
+	if h == m.highestCachedBlockHeight {
+		m.highestCachedBlockHeight = h - 1
```
omritoptix, line 43: same here . you're assuming all blocks in cache are by order.

block/retriever.go
```diff
@@ -116,6 +117,21 @@ func (m *Manager) ProcessNextDABatch(daMetaData *da.DASubmitMetaData) error {
...
+
+	lastBatch := batchResp.Batches[len(batchResp.Batches)-1]
+
+	if len(lastBatch.Blocks) == 0 {
+		m.logger.Info("no blocks in last batch from DA", "daHeight", daMetaData.Height)
+		return fmt.Errorf("no blocks in last batch from DA: %d", daMetaData.Height)
+	}
+
+	lastReceivedHeight := lastBatch.Blocks[len(lastBatch.Blocks)-1].Header.Height
+	types.LastReceivedDAHeight.Set(float64(lastReceivedHeight))
```
omritoptix, line 133: don't think all of this code is necessary.  also you're not guranteed here that the blocks are valid so I don't think that's good. I assume sergi means last valid block but I'm not sure if he means recieved or applied.
if applied, I'd simply say that in the apply block, you simply update the block based on it's source and call it the day. 
@srene what do you think?

srene: i think is better to get the da height from here, because if the same block is received from p2p its not going to be added to the cache and not applied from da, so this info will be lost. if the block is not valid i dont think is relevant in terms of debugging the syncing status.

block/block_cache.go
```diff
@@ -9,40 +9,50 @@ import (
...
+	types.BlockCacheSize.Set(float64(len(m.cache)))
+	types.LowestPendingBlockHeight.Set(float64(m.GetLowestCachedBlockHeight()))
+	types.HighestReceivedBlockHeight.Set(float64(m.GetHighestCachedBlockHeight()))
+}
+
+func (m *Cache) GetLowestCachedBlockHeight() float64 {
+	m.Lock()
+	defer m.Unlock()
+	var lowest uint64
+	for h := range m.cache {
```
omritoptix, line 38: why we iterate here vs just saving the lowest on first entrance and comparing each new block to it?

block/block_cache.go
```diff
@@ -9,40 +9,50 @@ import (
...
-		m.highestCachedBlockHeight = h - 1
-		types.HighestReceivedBlockHeight.Set(float64(m.highestCachedBlockHeight))
+	return float64(lowest)
+}
+
+func (m *Cache) GetHighestCachedBlockHeight() float64 {
+	m.Lock()
+	defer m.Unlock()
+	var highest uint64
+	for h := range m.cache {
```
omritoptix, line 50: same here. why we need to iterate?

omritoptix:
build failed

block/retriever.go
zale144: do we care if they are valid or not? I think we care if we received them or not

block/gossip.go
```diff
@@ -4,34 +4,31 @@ import (
...
 	block := eventData.Block
 	commit := eventData.Commit
-	m.retrieverMu.Lock() // needed to protect blockCache access
-	_, found := m.blockCache[block.Header.Height]
-	// It is not strictly necessary to return early, for correctness, but doing so helps us avoid mutex pressure and unnecessary repeated attempts to apply cached blocks
-	if found {
-		m.retrieverMu.Unlock()
+	height := block.Header.Height
+
+	types.LastReceivedP2PHeightGauge.Set(float64(height))
```
mtsitrin, line 20: I think it makes more sense to set it after checking if it's already cached

block/retriever.go
```diff
@@ -116,6 +117,21 @@ func (m *Manager) ProcessNextDABatch(daMetaData *da.DASubmitMetaData) error {
...
+
+	lastBatch := batchResp.Batches[len(batchResp.Batches)-1]
+
+	if len(lastBatch.Blocks) == 0 {
+		m.logger.Info("no blocks in last batch from DA", "daHeight", daMetaData.Height)
+		return fmt.Errorf("no blocks in last batch from DA: %d", daMetaData.Height)
+	}
+
+	lastReceivedHeight := lastBatch.Blocks[len(lastBatch.Blocks)-1].Header.Height
+	types.LastReceivedDAHeightGauge.Set(float64(lastReceivedHeight))
```
mtsitrin, line 133: don't we need to set it after a successful apply?

mtsitrin: nvm. saw @srene 's comment.
eventhough I still think metrics are not debug data, and from metric perspective, makes more sense to see the last block we **synced** from DA/SL

mtsitrin (approved):

block/block.go
```diff
@@ -21,7 +22,9 @@ func (m *Manager) applyBlock(block *types.Block, commit *types.Commit, blockMeta
 		return types.ErrInvalidBlockHeight
 	}
 
-	m.logger.Debug("Applying block", "height", block.Header.Height, "source", blockMetaData.source)
+	types.LastAppliedBlockSource.With(prometheus.Labels{"source": blockMetaData.source.String()}).Set(0)
+
```
danwt: imo should extract a function to metrics.go which takes a source as arg and handes 
```suggestion
		types.LastAppliedBlockSource.WithLabelValues(blockMetaData.source.String()).Set(0)
```

block/block_cache.go
```diff
@@ -0,0 +1,45 @@
...
+	m.Lock()
+	defer m.Unlock()
+	ret, found := m.cache[h]
+	return ret, found
+}
+
+func (m *Cache) HasBlockInCache(h uint64) bool {
+	_, found := m.GetBlockFromCache(h)
+	return found
+}
```
danwt, line 39: Could you rename the cache to BlockCache and the methods to Add,Delete,Get,Has?

block/block_cache.go
```diff
@@ -0,0 +1,45 @@
...
+	cache map[uint64]CachedBlock
+	sync.Mutex
+}
+
+func (m *Cache) AddBlockToCache(h uint64, b *types.Block, c *types.Commit) {
+	m.Lock()
+	defer m.Unlock()
+	m.cache[h] = CachedBlock{Block: b, Commit: c}
+	types.BlockCacheSizeGauge.Set(float64(len(m.cache)))
+}
```
danwt, line 19: func (m *Cache) AddBlockToCache(h uint64, b *types.Block, c *types.Commit) {
	m.Lock()
	m.cache[h] = CachedBlock{Block: b, Commit: c}
	m.Unlock()
	m.SetSizeMetric()
}

func (m *Cache) SetSizeMetric() {
	types.BlockCacheSizeGauge.Set(float64(m.Size()))
}

is preferable I think, you should reuse Size() if you implemented it

types/metrics.go
```diff
@@ -34,3 +34,37 @@ var RollappPendingSubmissionsSkewNumBytes = promauto.NewGauge(prometheus.GaugeOp
...
+
+var LastReceivedP2PHeightGauge = promauto.NewGauge(prometheus.GaugeOpts{
+	Name: "last_received_p2p_height",
+	Help: "The height of the last block received from P2P.",
+})
+
+var LastReceivedDAHeightGauge = promauto.NewGauge(prometheus.GaugeOpts{
+	Name: "last_received_da_height",
+	Help: "The height of the last block received from DA.",
+})
```
danwt, line 46: shouldn't it be one gauge with two dimensions? (p2p and da)

types/metrics.go
```diff
@@ -34,3 +34,37 @@ var RollappPendingSubmissionsSkewNumBytes = promauto.NewGauge(prometheus.GaugeOp
...
+		Name: "last_applied_block_source",
+		Help: "The source of the last applied block",
+	},
+	[]string{SourceLabel},
+)
+
+func SetLastAppliedBlockSource(source string) {
+	LastAppliedBlockSource.DeleteLabelValues("none", "p2p", "da")
+	LastAppliedBlockSource.With(prometheus.Labels{SourceLabel: source}).Set(0)
+}
```
danwt, line 65: WithLabelValues

block/types.go
```diff
@@ -9,17 +9,22 @@ import (
 
 // TODO: move to types package
 
-type blockSource string
+type BlockSource uint64
```
danwt, line 12: Starts with package name
should be Source

block/gossip.go
```diff
@@ -4,34 +4,31 @@ import (
...
+	m.logger.Debug("Received new block via gossip.", "block height", height, "store height", m.State.Height(), "n cachedBlocks", m.blockCache.Size())
 
 	nextHeight := m.State.NextHeight()
-	if block.Header.Height >= nextHeight {
-		m.blockCache[block.Header.Height] = CachedBlock{
-			Block:  &block,
-			Commit: &commit,
-		}
+	if height >= nextHeight {
+		m.blockCache.AddBlockToCache(height, &block, &commit)
```
danwt, line 30: seems to me that blocks can go to the cache and never get deleted if nextHeight increases in another thread

block/retriever.go
```diff
@@ -135,8 +137,12 @@ func (m *Manager) ProcessNextDABatch(daMetaData *da.DASubmitMetaData) error {
 				return fmt.Errorf("apply block: height: %d: %w", block.Header.Height, err)
 			}
 
-			delete(m.blockCache, block.Header.Height)
+			lastAppliedHeight = float64(block.Header.Height)
+
+			m.blockCache.DeleteBlockFromCache(block.Header.Height)
 		}
+
+		types.LastReceivedDAHeightGauge.Set(lastAppliedHeight)
```
danwt, line 145: Why not just 1 liner 
types.LastReceivedDAHeightGauge.Set( float64(block.Header.Height))

danwt (changes requested):
I can't reason why the cache cannot fill up with old blocks and leak memory.

You have two threads in onNewGossipedBlock and they both read nextHeight := m.State.NextHeight() and get  	if height >= nextHeight , but then one of them applies blocks
then the other adds a block which will never be deleted

block/block_cache.go
zale144: the package is called `block` already

block/types.go
zale144: moved to `types` package

block/gossip.go
zale144: removed block cache mutex

block/retriever.go
zale144: no need to set it every time a block is applied

mtsitrin (approved):

omritoptix (approved):

PR #991, opened 2024-07-30
feat: rollapp consensus params 

### LINKED ISSUE #989: params: add rollapp params checks
as described in https://www.notion.so/dymension/ADR-x-Rollapp-Consensus-Params-d60daee1f8cb42e38e43e1ef5289a220?pvs=4

srene:
# PR Standards

--

---

Close #989 

This PR refactors the rollapp params in dymint, according to the following ADR: https://www.notion.so/dymension/ADR-x-Rollapp-Consensus-Params-d60daee1f8cb42e38e43e1ef5289a220?pvs=4

The main modifications are the following:

New config checks:
- `max_batch_size`: Add validation `max_batch_size` is smaller than  max blob accepted for each DA.  celestia is 500KB.
- `max_batch_skew` : The `max_batch_skew` param is modified for max_block_skew, so the skew is measured in blocks and not in batches. the `max_block_skew` is configurable but with a limit set to 1 day of blocks at max block rate. 
- `max_batch_time`: Add validation is no more than 1 hour.
- `block_time` : check config is between 200 ms and 6 seconds

Existing consensus params:
- `max_block_size` : Add  limit produced block  `max_block_size`, which can never be greater than `max_batch_size`
-  update params on gov proposal

New consensus params:
- `rollapp version (commit)`: Add validation on startup for version specified in consensus param.
- `da_provider` : get it from consensus (genesis) instead of configurable.

New endblocker response to update new consensus params on gov proposal.

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling for block application processes.
	- Introduced `RollappConsensusParams` to improve consensus parameter management.
	- Added a new `GetMaxBlobSizeBytes` method to the Data Availability Layer client.
	- Streamlined initialization for clients and configuration management.
	- Added new constants for block processing time settings.
	- Removed deprecated fields and parameters from configuration structures.
	- Improved the handling of block production and submission processes.

- **Bug Fixes**
	- Improved error reporting and handling across multiple components.
	- Resolved issues with consensus parameter updates and state management.

- **Documentation**
	- Updated various files to reflect changes in configuration and consensus parameters.

- **Tests**
	- Enhanced testing framework with improved mock responses and restructuring of test cases.

- **Chores**
	- Updated module dependencies to reflect changes in underlying libraries.


omritoptix (changes requested):
My general comment is that I think the validation flow could be improved and simplified. What it comes to DA/version we have 3 types of variants which should check: 

1. What’s the actual version, da (from build and config)
2. What dymint has in state 
3. The new state (by the end blocker)

If (1),(2) match (which we can validate on startup), than you only need to check (1)=(3) or (2)=(3)

Currently you are validating (1) == (2) on startup and than (2) == (3) on change. I think it’s redundant and complicates things. How?   cause  if (2) != (3) now you need to first save the state and than handle the mismatch after the fact, which imo is unintuitive and creates a code smell ([look here](https://github.com/dymensionxyz/dymint/pull/991/files#diff-9a93a4d3f3ae40bcc564c225e9122a3d92f49c2a187b5da9f52715348766f788R90)).

Imo, have a generic function that validates (1) == (2). This function should be Called on:
1. Startup
2. Anytime after state update. 

If the function fails, than update your local version and than state update should work as usual.

proto/types/dymint/state.proto
```diff
@@ -38,4 +38,17 @@ message State {
 
   uint64 last_store_height = 16 [(gogoproto.customname) = "LastStoreHeight"];
   uint64 base_height = 17;
+  RollappConsensusParams consensus_params = 18 [(gogoproto.nullable) = false];
+
+}
+
+message RollappConsensusParams {
+  Params params = 1;
```
omritoptix, line 46: why do we need another `Params` message here vs just inlining the params ?

block/submit.go
```diff
@@ -35,14 +36,18 @@ func (m *Manager) SubmitLoop(ctx context.Context,
...
+	pendingSubmittedBlocks func() uint64,
 	maxBatchTime time.Duration, // max time to allow between batches
 	maxBatchBytes uint64, // max size of serialised batch in bytes
 	createAndSubmitBatch func(maxSizeBytes uint64) (sizeBlocksCommits uint64, err error),
 ) error {
 	eg, ctx := errgroup.WithContext(ctx)
 
 	pendingBytes := atomic.Uint64{}
+
+	pendingBlocks := atomic.Uint64{}
```
omritoptix, line 49: not sure I understand why we need to define an atomic value for it for saving, loading and  updating it vs just call `GetUnsubmittedBlocks` when necessary (i.e to just to check skew and possibly metrics)?

config/defaults.go
```diff
@@ -13,24 +13,22 @@ const (
...
-func DefaultConfig(home, chainId string) *NodeConfig {
+func DefaultConfig(home string) *NodeConfig {
 	cfg := &NodeConfig{
 		BlockManagerConfig: BlockManagerConfig{
 			BlockTime:          200 * time.Millisecond,
 			MaxIdleTime:        3600 * time.Second,
 			MaxProofTime:       100 * time.Second,
 			BatchSubmitMaxTime: 3600 * time.Second,
-			MaxBatchSkew:       20,
+			MaxBlockSkew:       90000,
```
omritoptix, line 29: why this value? seems like a lot.

config/config.go
```diff
@@ -18,6 +18,10 @@ const (
 	DefaultDymintDir      = ".dymint"
 	DefaultConfigDirName  = "config"
 	DefaultConfigFileName = "dymint.toml"
+	MinBlockTime          = 200 * time.Millisecond
+	MaxBlockTime          = 6 * time.Second
+	MaxBatchSubmitMaxTime = 1 * time.Hour
+	MaxBlockSkewSupported = 432000 // equivalent to 24h blocks at max block rate
```
omritoptix, line 24: seems way too much imo. 

what's the euristic for choosing this? how long will it take to recover from this? (i.e be in sync again with settlement, i.e submission rate to cover 432000.)

config/toml.go
```diff
@@ -70,7 +70,7 @@ block_time = "{{ .BlockManagerConfig.BlockTime }}"
 # block production interval in case of no transactions ("0s" produces empty blocks)
 max_idle_time = "{{ .BlockManagerConfig.MaxIdleTime }}"
 max_proof_time = "{{ .BlockManagerConfig.MaxProofTime }}"
-max_supported_batch_skew = {{ .BlockManagerConfig.MaxBatchSkew }}
+max_supported_block_skew = {{ .BlockManagerConfig.MaxBlockSkew }}
```
omritoptix, line 73: why is it called `max_supported_block_skew` (which is a system definition) if its a user defined param (i.e `max_block_skew`)

config/config.go
```diff
@@ -53,8 +56,8 @@ type BlockManagerConfig struct {
 	MaxProofTime time.Duration `mapstructure:"max_proof_time"`
 	// BatchSubmitMaxTime is how long should block manager wait for before submitting batch
 	BatchSubmitMaxTime time.Duration `mapstructure:"batch_submit_max_time"`
-	// MaxBatchSkew is the number of batches which are waiting to be submitted. Block production will be paused if this limit is reached.
-	MaxBatchSkew uint64 `mapstructure:"max_supported_batch_skew"`
+	// MaxBlockSkew is the number of blocks which are waiting to be submitted. Block production will be paused if this limit is reached.
+	MaxBlockSkew uint64 `mapstructure:"max_supported_block_skew"`
```
omritoptix, line 60: confusing naming as `max_supported_block_skew` is the max dymint support, not what the user chooses.

config/config.go
```diff
@@ -18,6 +18,10 @@ const (
 	DefaultDymintDir      = ".dymint"
 	DefaultConfigDirName  = "config"
 	DefaultConfigFileName = "dymint.toml"
+	MinBlockTime          = 200 * time.Millisecond
+	MaxBlockTime          = 6 * time.Second
+	MaxBatchSubmitMaxTime = 1 * time.Hour
```
omritoptix, line 23: why twice `Max` ?

omritoptix: I think in general we should save `max` to be system limits and without `max` to be user defined. e.g `MaxBatchSubmitTime` for the system and `BatchSubmitTime` for user. not sure if `max` makes sense in this aspect if the user is defining the param

config/config.go
```diff
@@ -153,12 +160,20 @@ func (c BlockManagerConfig) Validate() error {
...
 		return fmt.Errorf("block_batch_size_bytes must be positive")
 	}
 
-	if c.MaxBatchSkew <= 0 {
-		return fmt.Errorf("max_supported_batch_skew must be positive")
+	if c.MaxBlockSkew <= 0 {
+		return fmt.Errorf("max_supported_block_skew must be positive")
+	}
+
+	if c.MaxBlockSkew > MaxBlockSkewSupported {
```
omritoptix, line 175: should we also have a minumum for that?

config/config.go
```diff
@@ -153,12 +160,20 @@ func (c BlockManagerConfig) Validate() error {
 		return fmt.Errorf("batch_submit_max_time must be greater than max_idle_time")
 	}
 
+	if c.BatchSubmitMaxTime > MaxBatchSubmitMaxTime {
```
omritoptix, line 163: should we also have a mininum for that?

omritoptix (changes requested):

block/manager.go
```diff
@@ -108,11 +109,9 @@ func NewManager(
 		Genesis:          genesis,
 		Store:            store,
 		Executor:         exec,
-		DAClient:         dalc,
 		SLClient:         settlementClient,
-		Retriever:        dalc.(da.BatchRetriever),
 		targetSyncHeight: diodes.NewOneToOne(1, nil),
-		logger:           logger,
+		logger:           logger.With("module", "BlockManager"),
```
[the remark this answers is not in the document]
srene: done

node/node.go
```diff
@@ -201,6 +188,16 @@ func NewNode(
 		cancel:         cancel,
 	}
 
+	err = node.setDA(dalcKV, logger)
+	if err != nil {
+		return nil, err
+	}
+
+	err = blockManager.ValidateRollappParams()
```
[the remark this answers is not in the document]
srene: done

block/manager.go
```diff
@@ -239,3 +238,18 @@ func (m *Manager) UpdateTargetHeight(h uint64) {
 		}
 	}
 }
+
+func (m *Manager) ValidateRollappParams() error {
```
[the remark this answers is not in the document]
srene: this is actually only doing 2, but these config  params validation depends on consensus params and therefore it needs to have access to blockmanager State. i renamed the func to be more clear.

proto/types/tendermint/types/params.proto
```diff
@@ -78,3 +78,9 @@ message HashedParams {
   int64 block_max_bytes = 1;
   int64 block_max_gas   = 2;
 }
+
+
+message RollappConsensusParams {
```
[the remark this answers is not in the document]
srene: actually not used. removed

block/block.go
```diff
@@ -83,9 +84,10 @@ func (m *Manager) applyBlock(block *types.Block, commit *types.Commit, blockMeta
 	// If failed here, after the app committed, but before the state is updated, we'll update the state on
 	// UpdateStateFromApp using the saved responses and validators.
 
-	// Update the state with the new app hash, last validators and store height from the commit.
+	// Update the state with the new app hash, last validators, store height from the commit and rollapp params.
 	// Every one of those, if happens before commit, prevents us from re-executing the block in case failed during commit.
-	m.Executor.UpdateStateAfterCommit(m.State, responses, appHash, block.Header.Height, validators)
+	// If a rollapp param update requires halt, it returns an error, that will be used to stop the node afterwards.
```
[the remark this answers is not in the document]
srene: code updated, not relevant anymore

block/state.go
```diff
@@ -141,19 +125,37 @@ func (e *Executor) UpdateStateAfterInitChain(s *types.State, res *abci.ResponseI
...
+
+	if resp.EndBlock.RollappConsensusParamUpdates == nil {
+		return nil
+	}
+	s.ConsensusParams.Params.BlockMaxSize = resp.EndBlock.RollappConsensusParamUpdates.Block.MaxBytes
+	s.ConsensusParams.Params.BlockMaxGas = resp.EndBlock.RollappConsensusParamUpdates.Block.MaxGas
+
+	var err error
+	if s.ConsensusParams.Params.Da != resp.EndBlock.RollappConsensusParamUpdates.Da {
+		e.logger.Debug("Updating DA", "da", s.ConsensusParams.Params.Da, "newda", resp.EndBlock.RollappConsensusParamUpdates.Da)
```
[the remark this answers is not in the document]
srene: removed

block/state.go
```diff
@@ -141,19 +125,37 @@ func (e *Executor) UpdateStateAfterInitChain(s *types.State, res *abci.ResponseI
...
+		s.ConsensusParams.Params.Da = resp.EndBlock.RollappConsensusParamUpdates.Da
+		err = fmt.Errorf("%w, please update da config for %s", ErrDAUpgrade, s.ConsensusParams.Params.Da)
```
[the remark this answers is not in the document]
srene: replaced

block/errors.go
```diff
@@ -5,5 +5,6 @@ import "errors"
 var (
 	ErrNonRecoverable = errors.New("non recoverable")
 	ErrRecoverable    = errors.New("recoverable")
-	ErrWrongDA        = errors.New("wrong DA")
+	ErrDAUpgrade      = errors.New("da upgrade")
+	ErrVersionUpgrade = errors.New("version upgrade")
```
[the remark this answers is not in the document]
srene: updated

da/celestia/config.go
```diff
@@ -16,6 +16,7 @@ const (
 	namespaceVersion        = 0
 	DefaultGasPrices        = 0.1
 	defaultRpcRetryAttempts = 5
+	maxBlobSize             = 500000
```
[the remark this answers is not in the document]
srene: updated

da/celestia/celestia.go
```diff
@@ -637,3 +637,8 @@ func (c *DataAvailabilityLayerClient) sync(rpc *openrpc.Client) {
 		c.logger.Error("Waiting for Celestia data availability client to sync", "err", err)
 	}
 }
+
+// GetMaxBlobSize returns the maximum allowed blob size in the DA, used to check the max batch size configured
+func (d *DataAvailabilityLayerClient) GetMaxBlobSize() uint32 {
+	return maxBlobSize
```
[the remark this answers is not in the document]
srene: updated

types/serialization.go
```diff
@@ -295,7 +295,7 @@ func (s *State) FromProto(other *pb.State) error {
 	s.LastHeightConsensusParamsChanged = other.LastHeightConsensusParamsChanged
 	copy(s.LastResultsHash[:], other.LastResultsHash)
 	copy(s.AppHash[:], other.AppHash)
-
+	s.ConsensusParams = other.ConsensusParams
```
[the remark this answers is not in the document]
srene: removed

proto/types/dymint/state.proto
srene: it helps reading params from genesis state

proto/types/tendermint/abci/types.proto
```diff
@@ -374,6 +375,12 @@ message Evidence {
   int64 total_voting_power = 5;
 }
 
+message RollappConsensusParams {
```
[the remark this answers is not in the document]
srene: actually not used. removed

node/node.go
```diff
@@ -349,3 +346,24 @@ func (n *Node) startPrometheusServer() error {
 func (n *Node) GetBlockManagerHeight() uint64 {
 	return n.BlockManager.State.Height()
 }
+
+// setDA initializes DA client in blockmanager according to DA type set in genesis or stored in state
+func (n *Node) setDA(dalcKV store.KV, logger log.Logger) error {
+	da_layer := n.BlockManager.State.ConsensusParams.Params.Da
```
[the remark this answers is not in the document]
srene: updated

config/config.go
srene: i think having a minimum value equivalent to max_idle_time is enough.

config/config.go
srene: i renamed batchsubmittime and batchsubmitbytes params according to this.

config/config.go
srene: param renamed

node/node.go
```diff
@@ -349,3 +346,24 @@ func (n *Node) startPrometheusServer() error {
 func (n *Node) GetBlockManagerHeight() uint64 {
 	return n.BlockManager.State.Height()
 }
+
+// setDA initializes DA client in blockmanager according to DA type set in genesis or stored in state
+func (n *Node) setDA(dalcKV store.KV, logger log.Logger) error {
```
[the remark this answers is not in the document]
srene: moved

config/toml.go
srene: renamed

block/manager.go
```diff
@@ -243,3 +253,49 @@ func (m *Manager) UpdateTargetHeight(h uint64) {
...
+	if da.Client(m.State.ConsensusParams.Params.Da) != m.DAClient.GetClientType() {
+		return fmt.Errorf("da client mismatch. rollapp param: %s da configured: %s", m.DAClient.GetClientType(), m.State.ConsensusParams.Params.Da)
+	}
+
+	return nil
+}
+
+// setDA initializes DA client in blockmanager according to DA type set in genesis or stored in state
+func (m *Manager) setDA(daconfig string, dalcKV store.KV, logger log.Logger) error {
+	da_layer := m.State.ConsensusParams.Params.Da
```
omritoptix, line 284: why is it snake case?

block/manager.go
```diff
@@ -243,3 +253,49 @@ func (m *Manager) UpdateTargetHeight(h uint64) {
...
 	}
 }
+
+// ValidateConfigWithRollappParams checks the configuration params are consistent with the params in the dymint state (e.g. DA and version)
+func (m *Manager) ValidateConfigWithRollappParams() error {
+	if version.Commit != m.State.ConsensusParams.Params.Commit {
+		return fmt.Errorf("binary version mismatch. rollapp param: %s binary used:%s", version.Commit, m.State.ConsensusParams.Params.Commit)
+	}
+
+	if m.DAClient == nil {
```
omritoptix, line 263: why is this validated here? we're already validating this in the setDa.

block/manager.go
```diff
@@ -243,3 +253,49 @@ func (m *Manager) UpdateTargetHeight(h uint64) {
...
+		return nil
+	}
+
+	if m.DAClient.GetMaxBlobSizeBytes() != 0 && m.DAClient.GetMaxBlobSizeBytes() < uint32(m.Conf.BatchSubmitBytes) {
```
omritoptix, line 267: why are we validating this `m.DAClient.GetMaxBlobSizeBytes()` if it's a constant?

block/produce.go
```diff
@@ -186,10 +186,17 @@ func (m *Manager) produceBlock(allowEmpty bool) (*types.Block, *types.Commit, er
 		return nil, nil, fmt.Errorf("apply block: %w: %w", err, ErrNonRecoverable)
 	}
 
+	// validate configuration params and rollapp consensus params keep in line
+	err = m.ValidateConfigWithRollappParams()
+	if err != nil {
+		panic(err)
```
omritoptix, line 192: i think it's better to return here the error wrapping ErrUnrecoverable. why?
1. In general we handle errors in the higher level callee 
2. we don't really want to panic cause that will get us in a restart loop which won't help, we want to have the higher level caller catch it and emit unhealthy event.

block/retriever.go
```diff
@@ -103,6 +103,12 @@ func (m *Manager) applyLocalBlock(height uint64) error {
 	}
 	m.retrieverMu.Unlock()
 
+	// validate configuration params and rollapp consensus params keep in line
+	err = m.ValidateConfigWithRollappParams()
```
omritoptix, line 107: why not just have this logic as a first thing in the `applyBlock` vs putting it in 2 places?  (currently both in `applyLocalBlock` and `produceBlock`)

types/state.go
```diff
@@ -61,3 +63,22 @@ func (s *State) NextHeight() uint64 {
 	}
 	return s.Height() + 1
 }
+
+// LoadConsensusFromAppState load rollapp_params from genesis doc app_state to the dymint state
+func (s *State) LoadConsensusFromAppState(appState json.RawMessage) error {
```
omritoptix, line 68: imo name can be improved. I suggest `SetConsensusParamsFromGenesis`

I don't think where in genesis (in this case, under the `app_state` is interesting information)

Also I don't think that the `Load` is the interesting part about this fucntion.

types/state.go
```diff
@@ -61,3 +63,22 @@ func (s *State) NextHeight() uint64 {
 	}
 	return s.Height() + 1
 }
+
+// LoadConsensusFromAppState load rollapp_params from genesis doc app_state to the dymint state
```
omritoptix, line 67:
```suggestion
// SetConsensusParamsFromGenesis sets the rollapp consensus params from genesis
```

block/submit.go
```diff
@@ -40,22 +41,24 @@ func SubmitLoopInner(
 	ctx context.Context,
 	logger types.Logger,
 	bytesProduced chan int, // a channel of block and commit bytes produced
-	maxBatchSkew uint64, // max number of batches that submitter is allowed to have pending
+	maxBlockSkew uint64, // max number of blocks that submitter is allowed to have pending
+	pendingSubmittedBlocks func() uint64,
```
omritoptix, line 45: not sure I get the name here.
I think you mean `pendingSubmissionBlocks`? 

cause `pending` and `submitted` are the contrary. 

also why not just `pendingBlocks` ? we're in the context of submission here anyway.

block/submit.go
```diff
@@ -40,22 +41,24 @@ func SubmitLoopInner(
...
+
 	trigger := uchannel.NewNudger()   // used to avoid busy waiting (using cpu) on trigger thread
 	submitter := uchannel.NewNudger() // used to avoid busy waiting (using cpu) on submitter thread
 
 	eg.Go(func() error {
 		// 'trigger': this thread is responsible for waking up the submitter when a new block arrives, and back-pressures the block production loop
 		// if it gets too far ahead.
 		for {
-			if maxBatchSkew*maxBatchBytes < pendingBytes.Load() {
+			if maxBlockSkew <= pendingSubmittedBlocks() {
```
omritoptix, line 61: As I see it there is an inherent problem with this type of measurement of skew.
the reason being is that we need some way to correlate maxBatchSize with pending blocks. if I set maxBlockSkew to e.g 100, but all the blocks are e.g 1kb, than I get a skew before my batch is considered "full" (if my batch is 500kb).

so basically we're referring to the same issue in 2 different measures that needs to be coordinated, cause batch size defined may be > maxBlockSkew.

The problem with maxBlockSkew is that you can't really predict it's size as it depends on the block size. 

not sure why did you change it from the first place?

config/config.go
```diff
@@ -141,24 +148,33 @@ func (c BlockManagerConfig) Validate() error {
 		}
 	}
 
-	if c.BatchSubmitMaxTime <= 0 {
+	if c.BatchSubmitTime <= 0 {
 		return fmt.Errorf("batch_submit_max_time must be positive")
```
omritoptix, line 152:
```suggestion
		return fmt.Errorf("batch_submit_time must be positive")
```

config/config.go
```diff
@@ -141,24 +148,33 @@ func (c BlockManagerConfig) Validate() error {
...
 	}
 
-	if c.BatchSubmitMaxTime < c.BlockTime {
+	if c.BatchSubmitTime < c.BlockTime {
 		return fmt.Errorf("batch_submit_max_time must be greater than block_time")
```
omritoptix, line 156:
```suggestion
		return fmt.Errorf("batch_submit_time must be greater than block_time")
```

config/config.go
```diff
@@ -141,24 +148,33 @@ func (c BlockManagerConfig) Validate() error {
...
 	}
 
-	if c.BatchSubmitMaxTime < c.MaxIdleTime {
+	if c.BatchSubmitTime < c.MaxIdleTime {
 		return fmt.Errorf("batch_submit_max_time must be greater than max_idle_time")
```
omritoptix, line 160:
```suggestion
		return fmt.Errorf("batch_submit_time must be greater than max_idle_time")
```

config/config.go
```diff
@@ -52,11 +55,11 @@ type BlockManagerConfig struct {
 	// MaxProofTime defines the max time to be idle, if txs that requires proof were included in last block
 	MaxProofTime time.Duration `mapstructure:"max_proof_time"`
 	// BatchSubmitMaxTime is how long should block manager wait for before submitting batch
-	BatchSubmitMaxTime time.Duration `mapstructure:"batch_submit_max_time"`
-	// MaxBatchSkew is the number of batches which are waiting to be submitted. Block production will be paused if this limit is reached.
-	MaxBatchSkew uint64 `mapstructure:"max_supported_batch_skew"`
+	BatchSubmitTime time.Duration `mapstructure:"batch_submit_max_time"`
```
omritoptix, line 58:
```suggestion
	BatchSubmitTime time.Duration `mapstructure:"batch_submit_time"`
```

config/config.go
```diff
@@ -52,11 +55,11 @@ type BlockManagerConfig struct {
...
+	// BatchSkewBlocks is the number of blocks which are waiting to be submitted. Block production will be paused if this limit is reached.
+	BatchSkewBlocks uint64 `mapstructure:"batch_skew_blocks"`
```
omritoptix, line 60: In general not sure the idea of measuring skew by blocks is good (see my other comment). regradless I'd change the name to
```suggestion
	MaxBlockSkew uint64 `mapstructure:"max_block_skew"`
```

config/config.go
```diff
@@ -141,24 +148,33 @@ func (c BlockManagerConfig) Validate() error {
 		}
 	}
 
-	if c.BatchSubmitMaxTime <= 0 {
+	if c.BatchSubmitTime <= 0 {
```
omritoptix, line 151: In general would make the comparison which check the same variable if possible in one line and just tell the range it should be within vs doing each comparison with it's own error message.

omritoptix (changes requested):
Not sure about measuring the skew using block as as I see it, it can easily collide with batch size. wrote a more elaborated comment about it in the submission thread.

config/config.go
srene: rolled back to batch skew

config/config.go
srene: updated

config/config.go
srene: updated

config/config.go
srene: updated

block/manager.go
srene: i think if forgot to remove it after a refactor

block/manager.go
srene: changed

block/manager.go
srene: we have to validate BatchSubmitBytes, according to the max blob value for the specific DA, that can change if the DA is updated.

block/produce.go
srene: i removed the panic here, but then i added panic if error after applying in the p2p onReceivedBlock, otherwise the error will just be logged and can be missed.

block/submit_loop_test.go
```diff
@@ -26,21 +25,7 @@ type testArgs struct {
...
-	for range args.nParallel {
-		wg.Add(1)
-		go func() {
-			testSubmitLoopInner(t, args)
-			wg.Done()
-		}()
-	}
-	wg.Wait()
-}
-
```
danwt, line 42: why gone?

block/submit_loop_test.go
```diff
@@ -68,10 +54,7 @@ func testSubmitLoopInner(
 					return
 				default:
 				}
-				// producer shall not get too far ahead
-				absoluteMax := (args.batchSkew + 1) * args.batchBytes // +1 is because the producer is always blocked after the fact
-				nProduced := nProducedBytes.Load()
-				require.True(t, nProduced < absoluteMax, "produced bytes not less than maximum", "nProduced", nProduced, "max", absoluteMax)
```
danwt, line 74: what is being tested if this is removed?

danwt (changes requested):

block/manager.go
omritoptix: hmm. not sure I get it.
if you'll add another DA, you'll add another const (which will probably not be zero).
so still don't understand what does it have to do with adding a DA..

block/produce.go
omritoptix: yea so I see we call `applyBlock` (in case of full node):
1. from da 
2. from p2p

seems like we're panicing here, and don't go into unhealthy mode - though we probably should - as I don't see the benefit of constant panicing. 

@srene  @danwt wdyt.

block/submit_loop_test.go
srene: added it again

block/submit_loop_test.go
srene: fixed

da/errors.go
```diff
@@ -21,4 +21,8 @@ var (
 	ErrProofNotMatching = errors.New("proof not matching")
 	// ErrNameSpace is returned when wrong namespace used
 	ErrNameSpace = errors.New("namespace not matching")
+
+	ErrDANotMatching = errors.New("DA in config not matching DA path")
+
+	ErrNonexistentDA = errors.New("nonexistent DA")
```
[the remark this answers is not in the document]
srene: removed

da/errors.go
```diff
@@ -21,4 +21,8 @@ var (
 	ErrProofNotMatching = errors.New("proof not matching")
 	// ErrNameSpace is returned when wrong namespace used
 	ErrNameSpace = errors.New("namespace not matching")
+
+	ErrDANotMatching = errors.New("DA in config not matching DA path")
```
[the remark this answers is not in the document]
srene: updated

config/config.go
```diff
@@ -124,41 +126,37 @@ func (nc NodeConfig) Validate() error {
...
 	if c.MaxIdleTime < 0 {
 		return fmt.Errorf("max_idle_time must be positive or zero to disable")
 	}
 	// MaxIdleTime zero disables adaptive block production.
 	if c.MaxIdleTime != 0 {
-		if c.MaxIdleTime <= c.BlockTime {
-			return fmt.Errorf("max_idle_time must be greater than block_time")
+		if c.MaxIdleTime <= c.BlockTime || c.MaxIdleTime > MaxBatchSubmitTime {
+			return fmt.Errorf("max_idle_time must be greater than block_time and not greater than %s", MaxBatchSubmitTime)
 		}
```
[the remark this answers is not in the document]
srene: updated

block/manager.go
```diff
@@ -134,6 +139,12 @@ func (m *Manager) Start(ctx context.Context) error {
 		return err
 	}
 
+	// validate configuration params and rollapp consensus params are in line
+	err = m.ValidateConfigWithRollappParams()
+	if err != nil {
+		return err
+	}
+
```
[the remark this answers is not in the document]
srene: moved

block/block.go
```diff
@@ -97,6 +97,15 @@ func (m *Manager) applyBlock(block *types.Block, commit *types.Commit, blockMeta
...
 	}
+
+	m.blockCache.Delete(block.Header.Height)
+
+	// validate configuration params and rollapp consensus params keep in line
+	err = m.ValidateConfigWithRollappParams()
+	if err != nil {
+		return err
+	}
+
```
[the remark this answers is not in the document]
srene: its here to validate config because rollapp params can be updated after block execution. comment extended for clarification.

block/block.go
```diff
@@ -97,6 +97,15 @@ func (m *Manager) applyBlock(block *types.Block, commit *types.Commit, blockMeta
 			m.logger.Error("prune blocks", "retain_height", retainHeight, "err", err)
 		}
 	}
+
+	m.blockCache.Delete(block.Header.Height)
```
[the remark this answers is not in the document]
srene: exactly

proto/types/dymint/state.proto
```diff
@@ -38,4 +38,17 @@ message State {
...
 }
+
+message Params {
+  int64 block_max_gas = 1;
+  int64 block_max_size = 2;
+  string da = 3 ;
+  string commit = 4 ;
+}
```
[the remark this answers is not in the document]
srene: added

block/produce.go
danwt: makes sense 👍

srene: i replaced panic by health status event

block/manager.go
```diff
@@ -243,3 +253,45 @@ func (m *Manager) UpdateTargetHeight(h uint64) {
 		}
 	}
 }
+
+// ValidateConfigWithRollappParams checks the configuration params are consistent with the params in the dymint state (e.g. DA and version)
+func (m *Manager) ValidateConfigWithRollappParams() error {
```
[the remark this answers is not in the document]
danwt: nvm
i understand it now but it does require some context

block/submit_loop_test.go
```diff
@@ -103,11 +106,15 @@ func testSubmitLoopInner(
 		timeSinceLast := time.Since(timeLastProgressT).Milliseconds()
 		require.True(t, timeSinceLast < absoluteMax, "too long since last update", "timeSinceLast", timeSinceLast, "max", absoluteMax)
 
+		pendingBlocks.Store(0)                    // no pending blocks to be submitted
```
danwt, line 109: I think strictly, this should not be 0 every time you submit batch, but I guess it makes no difference as only used for logging

danwt (commented):
There are still unresolved and unresponded feedbacks from previous reviews

block/submit.go
```diff
@@ -69,9 +72,12 @@ func SubmitLoopInner(
 					return ctx.Err()
 				case n := <-bytesProduced:
 					pendingBytes.Add(uint64(n))
-					logger.Info("Added bytes produced to bytes pending submission counter.", "n", n)
+					logger.Debug("Added bytes produced to bytes pending submission counter.", "bytes added", n, "pending", pendingBytes.Load())
 				}
 			}
+
+			types.RollappPendingSubmissionsSkewBytes.Set(float64(pendingBytes.Load()))
+			types.RollappPendingSubmissionsSkewBlocks.Set(float64(pendingSubmittedBlocks()))
```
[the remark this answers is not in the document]
danwt: ?

block/block.go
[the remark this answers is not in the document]
danwt: ok I got it 👍

omritoptix (approved):

Issue #999, opened 2024-08-07
Creating empty batches causes submit loop to get stuck 


omritoptix:
```
Aug 06 15:38:59 ip-10-200-8-4 rollappd[1750340]: time="2024-08-06T15:38:59Z" level=info msg="Submitted batch to SL.[start height 7986946 end height 7987290]" module=BlockManager
Aug 06 15:39:09 ip-10-200-8-4 rollappd[1750340]: time="2024-08-06T15:39:09Z" level=info msg="Submitted batch to DA.[start height 7987291 end height 7987607]" module=BlockManager
Aug 06 15:39:10 ip-10-200-8-4 rollappd[1750340]: time="2024-08-06T15:39:10Z" level=info msg="Batch accepted.[startHeight 7987291 endHeight 7987607 stateIndex 52747]" module=settlement_client
Aug 06 15:39:10 ip-10-200-8-4 rollappd[1750340]: time="2024-08-06T15:39:10Z" level=info msg="Submitted batch to SL.[start height 7987291 end height 7987607]" module=BlockManager
Aug 06 15:39:10 ip-10-200-8-4 rollappd[1750340]: time="2024-08-06T15:39:10Z" level=error msg="Submit blob.[error blobs and commitments: blob data must be > 0 && <= 1974272, but it was 0 bytes]" module=celestia
```

danwt:
maybe related to https://github.com/dymensionxyz/dymint/issues/995

danwt:
Solved by a few prs 
https://github.com/dymensionxyz/dymint/pull/991#discussion_r1718662624
https://github.com/dymensionxyz/dymint/commit/d3b43117ced84efb7dd7c69beb0a0c9dc92ca171
https://github.com/dymensionxyz/dymint/commit/f3e69a36dc878786083531c10c5c8c9681c9ad11
https://github.com/dymensionxyz/dymint/commit/7ee8303601cd2290137804e7fcab7adfdbf91c38
https://github.com/dymensionxyz/dymint/commit/d31bcf1f2ad92e237ab663d1429d148887a71e40

PR #1000, opened 2024-08-07
fix(submit loop): add more logging around skew calculation

danwt:
Adds more visibility to try to fix https://github.com/dymensionxyz/dymint/issues/999 in a followup PR.

- Immediately log any errors from worker routines
- More logs and log information around pending bytes management
- More explicit error when trying to create a batch but last submitted height is equal to last block height
- Avoid a race condition on startup when calculation bytes produced and sending to channel
- Consolidate call sites of last height set/get

## Summary by CodeRabbit

- **New Features**
	- Introduced encapsulated methods for setting and retrieving block height, improving state management.
	- Added a new utility function for atomic subtraction to enhance thread safety.
	- Implemented a logging function for better error handling in concurrent operations.

- **Bug Fixes**
	- Enhanced error handling and logging within the Manager class's execution loops.

- **Tests**
	- Refined tests to align with the new block height management methods.
	- Introduced property-based testing for atomic operations.

- **Chores**
	- Updated project dependencies to include a new library for property-based testing.


omritoptix (dismissed):

danwt:
Had to fix test logger

utils/atomic/funcs.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"sync/atomic"
+)
+
+/*
+TODO: move to sdk-utils
+*/
+
+func Uint64Sub(x *atomic.Uint64, y uint64) uint64 {
```
zale144, line 11: can you add a comment mentioning that this uses two's complement arithmetic

utils/atomic/funcs.go
```diff
@@ -0,0 +1,13 @@
+package atomic
+
+import (
+	"sync/atomic"
+)
+
+/*
+TODO: move to sdk-utils
```
zale144, line 8: why not move it for this PR?

utils/atomic/funcs_test.go
```diff
@@ -0,0 +1,36 @@
...
+	_ = flag.Set("rapid.steps", "50")
+
+	rapid.Check(t, func(r *rapid.T) {
+		exp := uint64(0)
+		got := atomic.Uint64{}
+		r.Repeat(map[string]func(r *rapid.T){
+			"": func(r *rapid.T) {
+				require.Equal(t, exp, got.Load())
+			},
+			"add": func(r *rapid.T) {
```
zale144, line 23: why is this case called "add" if you are testing a subtraction function? Is this a _rapid_ thing?

utils/errors/err_group.go
```diff
@@ -0,0 +1,24 @@
...
+
+/*
+TODO: move to sdk-utils
+*/
+
+// ErrGroupGoLog calls eg.Go on the errgroup but it will log the error immediately when it occurs
+// instead of waiting for all goroutines in the group to finish first. This has the advantage of making sure all
+// errors are logged, not just the first one, and it is more immediate. Also, it is guaranteed, in case that
+// of the goroutines is not properly context aware.
+func ErrGroupGoLog(eg *errgroup.Group, logger types.Logger, fn func() error) {
```
zale144, line 16: this is nice, but what if we want to include some extra information in the log message, like those that we can only get locally in the `fn`'s body?

zale144 (changes requested):

utils/atomic/funcs.go
danwt: need to sort out versions

utils/atomic/funcs.go
danwt: implementation detail, why is that relevant

utils/atomic/funcs_test.go
danwt: no, you can add negative numbers

utils/errors/err_group.go
danwt: cross that bridge when we get to it

utils/atomic/funcs.go
zale144: because not everyone might be familiar with how your function works

zale144: also, once again, can you please leave the conversation resolving to the reviewer

danwt: TBH I didn't know what twos complement is so the comment wouldn't have helped me
I just did an example on paper 
Not sure what benefit you get from 'twos complement' because if you understand twos complement you don't need the comment
And if you don't understand twos complement you will have to work out the math anyway

utils/atomic/funcs_test.go
danwt: changed anyway

utils/atomic/funcs.go
zale144: I mean, some docstring would be nice

danwt: done

zale144: the expression ^(y - 1) flips all the bits of the result of (y - 1), which gives you -y
then you just need to do addition with that negative number. I assume you did it like this since atomic package doesn't have a Sub method

danwt: >the expression ^(y - 1) flips all the bits of the result of (y - 1), which gives you -y
then you just need to do addition with that negative number.

I understand that, this is just repeating the code in english

> I assume you did it like this since atomic package doesn't have a Sub method

👍

utils/atomic/funcs.go
```diff
@@ -0,0 +1,15 @@
...
+*/
+
+// Uint64Sub does x := x-y and returns the new value of x
+func Uint64Sub(x *atomic.Uint64, y uint64) uint64 {
+	// Uses math
```
zale144, line 13: 😆 might as well say `// Uses magic`

block/manager.go
```diff
@@ -164,27 +165,25 @@ func (m *Manager) Start(ctx context.Context) error {
...
-		eg.Go(func() error {
+		uerrors.ErrGroupGoLog(eg, m.logger, func() error {
 			return m.SyncToTargetHeightLoop(ctx)
 		})
 	}
 
 	go func() {
-		err := eg.Wait()
-		m.logger.Info("Block manager err group finished.", "err", err)
+		_ = eg.Wait() // errors are already logged
```
zale144, line 185: so instead of logging the error here, it is now logged inside `uerrors.ErrGroupGoLog`?
I might be missing something, but I don't see how we benefit with this change.

zale144: ah, nevermind. You wanted to log the error right away, not to wait for all the goroutines to end

zale144 (approved):

utils/atomic/funcs.go
danwt: might as well say nothing 😄

zale144: no, I like it as it is :)

omritoptix (approved):

PR #1035, opened 2024-08-22
fix(p2p): update height in blocksync loop

### LINKED ISSUE #1031: blocksync may download blocks already applied
in case da is faster than p2p

srene:
# PR Standards

This PR fixes an issue when DA is downloading blocks faster than blocksync, which may cause blocksync keeps downloading old blocks already applied. 
--

---

Close #1031 

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/p2p/client.go
+++ b/p2p/client.go
@@ -597,6 +597,11 @@ func (c *Client) retrieveBlockSyncLoop(ctx context.Context, msgHandler BlockSync
 
 				c.logger.Debug("Blocksync block received ", "height", h)
 				msgHandler(&block)
+				state, err := c.store.LoadState()
+				if err != nil {
+					return
+				}
+				h = max(h, state.NextHeight()-1)
 			}
 			c.blocksReceived.RemoveBlocksReceivedUpToHeight(state.NextHeight())
 		}


omritoptix (approved):

omritoptix (approved):

p2p/client.go
```diff
@@ -597,6 +597,11 @@ func (c *Client) retrieveBlockSyncLoop(ctx context.Context, msgHandler BlockSync
 
 				c.logger.Debug("Blocksync block received ", "height", h)
 				msgHandler(&block)
+				state, err := c.store.LoadState()
+				if err != nil {
+					return
+				}
+				h = max(h, state.NextHeight()-1)
```
mtsitrin, line 604: why not setting it in the `for` loop definition?
also need to validate that `h` progress, otherwise we can be stuck in endless loop (in case the height doesn't progress)

srene: i think is fine, because the max func will cause that h is always progressing, it will just skip already applied heights. in case state.NextHeight() is updated in parallel

PR #1042, opened 2024-08-26
fix(manager): use block params from consensus param 

### LINKED ISSUE #1041: convert rollappparams block max gas to int64

### LINKED ISSUE #1064: use block params only from consensus params
instead of using max gas from consensus params and max bytes from rollapp params

srene:
# PR Standards

This PR refactors the rollapp params in dymint so max bytes is used again from consensus params + rollapp binary version field is renamed from commit to version
--

---

Close #1064

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


block/manager.go
```diff
@@ -328,3 +327,14 @@ func (m *Manager) setDA(daconfig string, dalcKV store.KV, logger log.Logger) err
 	m.Retriever = retriever
 	return nil
 }
+
+func (m *Manager) ValidateConsensusBlockParams() {
```
omritoptix, line 331: IMO the validation here is not necessary because:
1. we already have a max protection when creating the block (i.e min(maxSize, blockBytes))
2. The valiation should be done in the sdk level when voting on gov and putting it here imo is dirty an changing the state regardless to the actual gov could be confusing

block/executor.go
```diff
@@ -89,11 +89,9 @@ func (e *Executor) InitChain(genesis *tmtypes.GenesisDoc, valset []*tmtypes.Vali
 }
 
 // CreateBlock reaps transactions from mempool and builds a block.
-func (e *Executor) CreateBlock(height uint64, lastCommit *types.Commit, lastHeaderHash, nextSeqHash [32]byte, state *types.State, maxBlockDataSizeBytes uint64) *types.Block {
-	if state.ConsensusParams.Blockmaxsize > 0 {
-		maxBlockDataSizeBytes = min(maxBlockDataSizeBytes, uint64(state.ConsensusParams.Blockmaxsize))
-	}
-	mempoolTxs := e.mempool.ReapMaxBytesMaxGas(int64(maxBlockDataSizeBytes), state.ConsensusParams.Blockmaxgas)
+func (e *Executor) CreateBlock(height uint64, lastCommit *types.Commit, lastHeaderHash, nextSeqHash [32]byte, state *types.State, maxBlockDataSizeBytes uint32) *types.Block {
+	maxBlockDataSizeBytes = min(maxBlockDataSizeBytes, uint32(state.ConsensusParams.Block.MaxBytes))
```
omritoptix, line 93: if we want to add min protection here (so max (minsize, maxBytes) - than we can do it (instead of the validation function))

block/state.go
```diff
@@ -100,14 +102,34 @@ func (e *Executor) UpdateStateAfterInitChain(s *types.State, res *abci.ResponseI
...
 		copy(s.AppHash[:], res.AppHash)
 	}
-
+	if res.ConsensusParams != nil {
+		params := res.ConsensusParams
+		if params.Block != nil {
+			s.ConsensusParams.Block.MaxBytes = params.Block.MaxBytes
+			s.ConsensusParams.Block.MaxGas = params.Block.MaxGas
+		}
+		if params.Evidence != nil {
```
omritoptix, line 111: do we need the evidence params update?

block/state.go
```diff
@@ -100,14 +102,34 @@ func (e *Executor) UpdateStateAfterInitChain(s *types.State, res *abci.ResponseI
...
+			s.ConsensusParams.Evidence.MaxAgeNumBlocks = params.Evidence.MaxAgeNumBlocks
+			s.ConsensusParams.Evidence.MaxAgeDuration = params.Evidence.MaxAgeDuration
+			s.ConsensusParams.Evidence.MaxBytes = params.Evidence.MaxBytes
+		}
+		if params.Validator != nil {
+			// Copy params.Validator.PubkeyTypes, and set result's value to the copy.
+			// This avoids having to initialize the slice to 0 values, and then write to it again.
+			s.ConsensusParams.Validator.PubKeyTypes = append([]string{}, params.Validator.PubKeyTypes...)
```
omritoptix, line 119: do we need this?

block/state.go
```diff
@@ -100,14 +102,34 @@ func (e *Executor) UpdateStateAfterInitChain(s *types.State, res *abci.ResponseI
...
+		}
+		if params.Version != nil {
+			s.ConsensusParams.Version.AppVersion = params.Version.AppVersion
```
omritoptix, line 122: this?

omritoptix (changes requested):

block/executor.go
```diff
@@ -60,37 +62,20 @@ func (e *Executor) InitChain(genesis *tmtypes.GenesisDoc, valset []*tmtypes.Vali
...
-			Version: &tmproto.VersionParams{
-				AppVersion: params.Version.AppVersion,
-			},
-		},
-		Validators:    valUpdates,
-		AppStateBytes: genesis.AppState,
-		InitialHeight: genesis.InitialHeight,
+		Time:            genesis.GenesisTime,
+		ChainId:         genesis.ChainID,
+		ConsensusParams: &abci.ConsensusParams{},
```
omritoptix: why we removed this? not sure I follow. 
how can the app pass back after the block size etc if we're not initializing it from genesis with it?

block/executor.go
srene: removed validation and added  min protection

block/manager.go
srene: removed validation

block/state.go
srene: removed

block/state.go
srene: removed

block/state.go
srene: removed

block/executor.go
srene: readded

omritoptix (approved):

PR #1045, opened 2024-08-27
fix(manager): gossip any pending blocks not gossiped before (in case of restart or crash)

### LINKED ISSUE #1044: block is not found using blocksync after crash
because the block is applied from local db and not gossiped, and blocksync is using only gossiped blocks

srene:
# PR Standards

--

---

In case the node crashes there maybe blocks that are not gossiped, but applied. In this case, blocks will never be gossiped (and neither added to blocksync store) and therefore not retrievable from P2P. 
This pr fixes it by adding blocks to blocksync right after are produced and stored to db, this way blocks will be aways available in p2p, even if not gossiped.

Close #1044 

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


block/manager.go
```diff
@@ -193,6 +193,8 @@ func (m *Manager) Start(ctx context.Context) error {
 		m.handleRotationReq(ctx, nextSeqAddr)
 		return nil
 	}
+	// gossip any pending blocks that could not be gossiped after produced before node was stopped or crashing
```
omritoptix, line 196:
```suggestion
	// gossip any blocks that weren't gossiped before restart
```

block/manager.go
```diff
@@ -328,3 +330,22 @@ func (m *Manager) setDA(daconfig string, dalcKV store.KV, logger log.Logger) err
 	m.Retriever = retriever
 	return nil
 }
+
+func (m *Manager) gossipPendingBlocks(ctx context.Context) {
+	for h := m.State.LastGossipedHeight + 1; h <= m.State.Height(); h++ {
```
omritoptix, line 335: after upgrade the LastGossipedHiehgt will be zero no? so it will gossip all blocks.

block/manager.go
```diff
@@ -328,3 +330,22 @@ func (m *Manager) setDA(daconfig string, dalcKV store.KV, logger log.Logger) err
 	m.Retriever = retriever
 	return nil
 }
+
+func (m *Manager) gossipPendingBlocks(ctx context.Context) {
```
omritoptix, line 334: seems like we're not updating the LastGossipedHeight here or in the `gossipBlock` function.

block/produce.go
```diff
@@ -120,6 +120,12 @@ func (m *Manager) produceApplyGossip(ctx context.Context, allowEmpty bool, nextP
 		return nil, nil, fmt.Errorf("gossip block: %w", err)
 	}
 
+	m.State.LastGossipedHeight = block.Header.Height
```
omritoptix, line 123: why updating it here and not in the `gossipBlock` function? that way seems much less error-prone to have one place (cause now for example in the node recovery function for gossiping were not updating it)

omritoptix (changes requested):

block/manager.go
srene: updated in gossipBlock

block/produce.go
srene: moved inside function

block/manager.go
srene: true. i set it to the node height in case its 0 in the init. this way it will not start from zero when the upgrade.

block/manager.go
```diff
@@ -328,3 +330,22 @@ func (m *Manager) setDA(daconfig string, dalcKV store.KV, logger log.Logger) err
 	m.Retriever = retriever
 	return nil
 }
+
```
omritoptix, line 333: another edge case I see here is the following: I was a sequencer previously, and than I wasn't (rotation. but continued being a full node), and than I am again. In this case, my store `LastGossipedHeight` wasn't updated when I was a full node (or at least I don't see it). and than I need to get stuck on gossiping potentially a big amount. 

I think we need some sort of a sane skew after which we avoid gossiping (i.e if I'm 1M blocks away from the latest state, than probably it doesn't make sense).

block/state.go
```diff
@@ -32,6 +32,10 @@ func (m *Manager) LoadStateOnInit(store store.Store, genesis *tmtypes.GenesisDoc
 	}
 
 	m.State = s
+
+	if m.State.LastGossipedHeight == 0 {
```
omritoptix, line 36: worth adding a comment here explainig the reasoning

srene: comment added

block/manager.go
srene: i added an update LastGossipedHeight inside UpdateTargetHeight, this will cause full-nodes also update LastGossipedHeight when retrieving form p2p or DA. this way, when rotating the value will be updated

block/manager.go
```diff
@@ -285,6 +287,15 @@ func (m *Manager) UpdateTargetHeight(h uint64) {
 			break
 		}
 	}
+
+	// updating LastGossipedHeight to avoid gossiping long queue of blocks when sequencer rotation
+	if m.State.LastGossipedHeight < h {
```
omritoptix, line 292: 2 issues I have with that:
1. the function is called update target height, but it actually also updates last gossiped height
2. the param is called last gossiped height for a full node, but it's tracking actually a potential sequencer only metric even if you're not a sequencer which may never be relevant for you. so maybe we need to make it a bit more make sense cause it feels hacky now (as if one would see that he would think it's the last height the full node got from gossiping but it's not)
3. we call the store save state on each iteration vs just updating it once. not sure why we need to iterate vs just saving h? it's just overriding in a loop.

srene: in order to avoid these issues, i refactored the pr and i implemented a new and simpler approach.
the new approach relies the block is added to blocksync after is produced but not on gossip. 
this way if the block is not gossiped because a panic, it will be available through blocksync, and full-nodes will be able to sync via p2p anyway

block/block.go
```diff
@@ -48,6 +50,16 @@ func (m *Manager) applyBlock(block *types.Block, commit *types.Commit, blockMeta
 			return fmt.Errorf("save block: %w", err)
 		}
 
+		gossipedBlock := p2p.BlockData{Block: *block, Commit: *commit}
```
omritoptix, line 53: for cleanliness I'd suggest moving this code part to a `saveP2PBlock` method or some other name which makes sense an add a godoc to this function. 

also we why use `context.TODO()`? s

srene: moved to new func

p2p/client.go
```diff
@@ -519,6 +519,11 @@ func (c *Client) blockGossipReceived(ctx context.Context, block []byte) {
 		c.logger.Error("Publishing event.", "err", err)
 	}
 	if c.conf.BlockSyncEnabled {
+		_, err := c.store.LoadBlockCid(gossipedBlock.Block.Header.Height)
```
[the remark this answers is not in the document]
srene: yes, this is just to avoid storing the  block  twice if already stored.

omritoptix (approved):

Issue #1070, opened 2024-09-09
Detect: sequencer does wrong state transition according to STF - impl


omritoptix:
based on this ADR: https://www.notion.so/ADR-Fraud-handling-General-STF-Timestamp-8dbedb4cd4394032aa2e7aa9f3bfac4e?pvs=4

PR #1071, opened 2024-09-11
feat: block production misbehavior detection 

faultytolly:
# PR Standards

https://github.com/dymensionxyz/dymint/issues/1070

~https://www.notion.so/dymension/ADR-x-generic-Dymint-fraud-handling-7425a9ef065b458c989404c42be6a6cd~
~https://www.notion.so/dymension/ADR-x-STF-fraud-detection-105a4a51f86a80f88225da364d798ddd~
~https://www.notion.so/dymension/ADR-x-timestamp-fraud-detection-105a4a51f86a80478c68d385b254459e~

https://www.notion.so/dymension/ADR-L2-Block-Production-Misbehavior-10da4a51f86a806ab08cf0a4e1cecc37

--

---

Close #XXX

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


danwt:
Please link the adr(s) and issue(s)

types/validation_test.go
```diff
@@ -0,0 +1,213 @@
...
+			if tt.wantErr {
+				assert.Error(t, err)
+				assert.Contains(t, err.Error(), tt.errMsg)
+				if tt.isFraud {
+					require.True(t, errors.Is(err, fraud.ErrFraud))
+				} else {
+					require.False(t, errors.Is(err, fraud.ErrFraud))
+				}
+			} else {
+				assert.NoError(t, err)
```
danwt, line 209: Instead of testing using string compare you should remove errMsg and just use errors.Is with the expected type of fraud

types/errors.go
```diff
@@ -15,3 +17,81 @@ var (
...
 )
+
+type ErrFraudHeightMismatch struct {
+	expected uint64
+	actual   uint64
+
+	headerHeight uint64
+	headerHash   [32]byte
+	proposer     []byte
+}
```
danwt, line 28: shouldn't all of the fields on errors by public?

block/executor.go
```diff
@@ -19,6 +19,18 @@ import (
 // default minimum block max size allowed. not specific reason to set it to 10K, but we need to avoid no transactions can be included in a block.
 const minBlockMaxBytes = 10000
 
+type ExecutorI interface {
```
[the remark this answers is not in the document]
danwt: nvm it is already

danwt (changes requested):
Looking good

Aside from comments on the code, the main thing here 

- please over document everything wrt. security implications
- what is envisioned for fraud handler?
- dependence on accuracy of proposer addr

The fraud handler is going to need to check that everything is actually a fraud, rather than operator mistake (e.g. in wrong app-hash case) and that the fraud came from a particular slashable person. How do you envision that given the currently supplied data?

fraud/types.go
```diff
@@ -0,0 +1,7 @@
+package fraud
+
+import (
+	"errors"
+)
+
+var ErrFraud = errors.New("fraud error")
```
danwt, line 7: please use gerrc.ErrFault as a basis (wrap it)

block/retriever.go
```diff
@@ -146,7 +143,8 @@ func (m *Manager) ProcessNextDABatch(daMetaData *da.DASubmitMetaData) error {
 				continue
 			}
 
-			err := m.applyBlock(block, batch.Commits[i], types.BlockMetaData{Source: types.DA, DAHeight: daMetaData.Height})
+			// We dont validate because validateBlockBeforeApply already checks if the block is already applied, and we don't need to fail there.
+			err := m.applyBlockWithFraudHandling(block, batch.Commits[i], types.BlockMetaData{Source: types.DA, DAHeight: daMetaData.Height}, false)
```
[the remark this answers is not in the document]
srene: or remove the validate param in the function since it will be always validating then...

block/block.go
```diff
@@ -1,13 +1,42 @@
...
 
+// applyBlockWithFraudHandling calls applyBlock and validateBlockBeforeApply with fraud handling.
+func (m *Manager) applyBlockWithFraudHandling(block *types.Block, commit *types.Commit, blockMetaData types.BlockMetaData, validate bool) error {
+	if validate {
+		if err := m.validateBlockBeforeApply(block, commit); err != nil {
+			if errors.Is(err, fraud.ErrFraud) {
+				m.FraudHandler.HandleFault(context.Background(), err)
+			} else if err != nil {
+				m.blockCache.Delete(block.Header.Height)
+				// TODO: can we take an action here such as dropping the peer / reducing their reputation?
```
srene, line 23: i guess that would be just in case the proposer signature validation fails.

block/block.go
```diff
@@ -1,13 +1,42 @@
...
+			}
+
+			return fmt.Errorf("block not valid at height %d, dropping it: err:%w", block.Header.Height, err)
+		}
+	}
+
+	if err := m.applyBlock(block, commit, blockMetaData); err != nil {
+		if errors.Is(err, fraud.ErrFraud) {
```
srene, line 31: but applyBlock can also return ErrFraud? is it not only in validateBlockBeforeApply?

danwt (changes requested):
Just couple things

not sure if timestamp fraud will ever get caught if it's in validate basic, do all validate basic calls have the fraud err check wrapper?

block/block.go
danwt: but isn't consensus messages fraud the same concept as app hash fraud?

block/state.go
```diff
@@ -118,9 +123,10 @@ func (e *Executor) UpdateMempoolAfterInitChain(s *types.State) {
 }
 
 // UpdateStateAfterCommit updates the state with the app hash and last results hash
-func (e *Executor) UpdateStateAfterCommit(s *types.State, resp *tmstate.ABCIResponses, appHash []byte, height uint64) {
+func (e *Executor) UpdateStateAfterCommit(s *types.State, resp *tmstate.ABCIResponses, appHash []byte, height uint64, lastHeaderHash [32]byte) {
 	copy(s.AppHash[:], appHash[:])
 	copy(s.LastResultsHash[:], tmtypes.NewResults(resp.DeliverTxs).Hash())
+	copy(s.LastHeaderHash[:], lastHeaderHash[:])
```
srene, line 129: i think LastHeaderHash needs to be added as well in state serialization (types/serialization.go), otherwise it won't be stored and can be set to 0 in restart.

block/retriever.go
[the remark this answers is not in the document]
srene: i think if it fails to validate here it should not continue and notify fraud, in the same way it is done when applying block from p2p. therefore imo it is not necessary to have the validateBlockBeforeApply  call before applyBlockWithFraudHandling, since it could be called  inside the applyBlockWithFraudHandling (same way as p2p). this way the validate param in  applyBlockWithFraudHandling function is not really necessary, because it wont be used.

block/block.go
```diff
@@ -1,13 +1,42 @@
...
 	"github.com/dymensionxyz/dymint/types"
 )
 
+// applyBlockWithFraudHandling calls applyBlock and validateBlockBeforeApply with fraud handling.
+func (m *Manager) applyBlockWithFraudHandling(block *types.Block, commit *types.Commit, blockMetaData types.BlockMetaData, validate bool) error {
+	if validate {
+		if err := m.validateBlockBeforeApply(block, commit); err != nil {
+			if errors.Is(err, gerrc.ErrFault) {
+				m.FraudHandler.HandleFault(context.Background(), err)
+			} else if err != nil {
```
srene, line 21: not sure if there exists the case where it returns an error that is actually not a fraud.

types/state.go
```diff
@@ -42,6 +42,9 @@ type State struct {
 
 	// New rollapp parameters .
 	RollappParams dymint.RollappParams
+
+	// LastHeaderHash is the hash of the last block header.
+	LastHeaderHash [32]byte
```
mtsitrin, line 47: if it needs to be persistent, needs to be added to `serialization.go`

block/state.go
```diff
@@ -88,8 +88,13 @@ func (m *Manager) UpdateStateFromApp() error {
 		return errorsmod.Wrap(err, "load block responses")
 	}
 
+	block, err := m.Store.LoadBlock(appHeight)
```
mtsitrin, line 91: block already loaded by caller. u can pass `block.Header.Hash()` as arg

block/manager.go
```diff
@@ -125,7 +126,8 @@ func NewManager(
 		blockCache: &Cache{
 			cache: make(map[uint64]types.CachedBlock),
 		},
-		pruningC: make(chan int64, 10), // use of buffered channel to avoid blocking applyBlock thread. In case channel is full, pruning will be skipped, but the retain height can be pruned in the next iteration.
+		FraudHandler: nil,                  // TODO: create a default handler
```
mtsitrin, line 129: it's required to fix, otherwise `applyBlockWithFraudHandling` will panic on fraud

block/block.go
mtsitrin, line 21: isnt the deletion from blockCache shold be done by caller?
`applyBlockWithFraudHandling` can be called from different entry points

mtsitrin (changes requested):

types/validation.go
```diff
@@ -46,22 +47,46 @@ func (b *Block) ValidateBasic() error {
...
+	nextHeight := state.NextHeight()
+	if b.Header.Height != nextHeight {
+		return NewErrFraudHeightMismatch(state.NextHeight(), b.Header.Height, b)
 	}
 
-	if b.Header.Height != state.NextHeight() {
-		return errors.New("height mismatch")
+	if !bytes.Equal(b.Header.NextSequencersHash[:], state.Sequencers.ProposerHash()) { // will change when sequencer rotation is implemented
+		return NewErrInvalidNextSequencersHashFraud([32]byte(state.Sequencers.ProposerHash()), b.Header.NextSequencersHash)
 	}
```
danwt, line 82: sequencer rotation is already implemented, is something unfinished?

types/validation.go
```diff
@@ -97,16 +122,35 @@ func (c *Commit) ValidateBasic() error {
 
 func (c *Commit) ValidateWithHeader(proposerPubKey tmcrypto.PubKey, header *Header) error {
 	if err := c.ValidateBasic(); err != nil {
-		return err
+		return NewErrInvalidSignatureFraud(err)
+	}
+
+	if c.Height != header.Height {
+		return NewErrInvalidBlockHeightFraud(c.Height, header.Height)
+	}
```
danwt, line 130: invalid block height fraud is a bit undefined IMO
this is an invalid commit fraud

types/validation.go
```diff
@@ -97,16 +122,35 @@ func (c *Commit) ValidateBasic() error {
 
 func (c *Commit) ValidateWithHeader(proposerPubKey tmcrypto.PubKey, header *Header) error {
 	if err := c.ValidateBasic(); err != nil {
-		return err
+		return NewErrInvalidSignatureFraud(err)
```
danwt, line 125: none of the frauds returned here will be caught in p2p case right? because there is no handler
https://github.com/dymensionxyz/dymint/blob/88ba1fee37359b65a1faf6de32fa94f2ae0c5330/p2p/validator.go#L83-L86

types/errors.go
```diff
@@ -14,4 +16,298 @@ var (
...
+
+func (e ErrTimeFraud) Unwrap() error {
+	return gerrc.ErrFault
+}
+
+type ErrLastHeaderHashMismatch struct {
+	Expected       [32]byte
+	LastHeaderHash [32]byte
+}
```
danwt, line 142: don't you need more info to validate this

types/errors.go
```diff
@@ -14,4 +16,298 @@ var (
...
+	}
+}
+
+func (e ErrInvalidBlockHeightFraud) Error() string {
+	return fmt.Sprintf("invalid block height. expected=%d, got=%d", e.Expected, e.ActualHeight)
+}
+
+func (e ErrInvalidBlockHeightFraud) Unwrap() error {
+	return gerrc.ErrFault
+}
```
danwt, line 197: can you document how this is different from the height mismatch fraud?

types/errors.go
```diff
@@ -14,4 +16,298 @@ var (
...
+	return fmt.Sprintf("invalid next sequencers hash. expected=%X, got=%X", e.ExpectedHash, e.ActualHash)
+}
+
+func (e ErrInvalidNextSequencersHashFraud) Unwrap() error {
+	return gerrc.ErrFault
+}
+
+type ErrInvalidHeaderDataHashFraud struct {
+	Err error
+}
```
danwt, line 299: what data is in here?

danwt (changes requested):
Looks like the scope has drifted a bit from what I originally understood

I've posted on slack to get this confirmed

https://dymensionworkspace.slack.com/archives/C07HV4329LJ/p1729509868648489

types/validation.go
```diff
@@ -97,16 +122,35 @@ func (c *Commit) ValidateBasic() error {
...
 	abciHeaderPb := ToABCIHeaderPB(header)
 	abciHeaderBytes, err := abciHeaderPb.Marshal()
 	if err != nil {
 		return err
 	}
 	// commit is validated to have single signature
 	if !proposerPubKey.VerifySignature(abciHeaderBytes, c.Signatures[0]) {
-		return ErrInvalidSignature
+		return NewErrInvalidSignatureFraud(ErrInvalidSignature)
 	}
```
danwt, line 154: doesn't this need to come before all other checks?

danwt (approved):

srene (approved):

PR #1072, opened 2024-09-13
fix(store): set optimal badger config to avoid memory spikes

### LINKED ISSUE #1067: Badger compaction memory optimization

srene:
# PR Standards

--

---

Close #1067 

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


store/badger.go
```diff
@@ -1,12 +1,20 @@
...
 	"github.com/dymensionxyz/gerr-cosmos/gerrc"
 
-	"github.com/dgraph-io/badger/v3"
+	"github.com/dgraph-io/badger/v4"
+	"github.com/dgraph-io/badger/v4/options"
+)
+
+const (
+	gcTimeout    = 1 * time.Minute
+	discardRatio = 0.5
```
danwt, line 17: could you comment a justification for these? link to some relevant forum thread or reasoning?

store/badger.go
```diff
@@ -59,6 +71,24 @@ func (b *BadgerKV) Close() error {
...
 
+func (b *BadgerKV) gc(period time.Duration, discardRatio float64) {
+	gcTimeout := time.NewTimer(period)
+	defer gcTimeout.Stop()
+
+	ctx := context.Background()
+	for {
+		select {
+		case <-ctx.Done():
+			return
```
danwt, line 82: ctx is useless

store/badger.go
```diff
@@ -59,6 +71,24 @@ func (b *BadgerKV) Close() error {
...
+		case <-gcTimeout.C:
+			err := b.db.RunValueLogGC(discardRatio)
+			if err != nil {
+				continue
```
danwt, line 86: better to log?

danwt (changes requested):

store/badger.go
srene: done

store/badger.go
srene: log added

store/badger.go
```diff
@@ -30,20 +40,25 @@ func NewDefaultInMemoryKVStore() KV {
...
+	db, err := badger.Open(*opts)
 	if err != nil {
 		panic(err)
 	}
-	return &BadgerKV{
+	b := &BadgerKV{
 		db: db,
 	}
+	if ctx != context.TODO() {
+		go b.gc(ctx, gcTimeout, discardRatio, logger)
```
danwt, line 54: bit weird
can you comment?

danwt: what exactl is the problem with running the gc even with context todo?
Cant you just handle ErrGCInMemoryMode at the callsite?

danwt (changes requested):
Looks good just not sure about this conditional code based on context

store/badger.go
srene: i think it does not make sense. i actually removed the context.

danwt (approved):

danwt (approved):

store/badger.go
```diff
@@ -59,6 +71,18 @@ func (b *BadgerKV) Close() error {
 	return b.db.Close()
 }
 
+func (b *BadgerKV) gc(period time.Duration, discardRatio float64, logger types.Logger) {
+	ticker := time.NewTicker(period)
+	defer ticker.Stop()
+	for range ticker.C {
```
mtsitrin, line 77: add graceful shutdown handling
and probably godoc
```
case <-db.ctx.Done():
    // Exit the function if the database context is cancelled
    return
}

File: store/badger.go [hunk 6]
```diff
@@ -189,3 +213,36 @@ func (i *BadgerIterator) Discard() {
...
+	// SyncWrites is a configuration option in Badger that determines whether writes are immediately synced to disk or no.
+	// If set to true it writes to the write-ahead log (value log) are synced to disk before being applied to the LSM tree.
+	opts.SyncWrites = syncWrites
+	// default 64mib => 0 - disable block cache
+	// BlockCacheSize specifies how much data cache should hold in memory.
+	// It improves lookup performance but increases memory consumption.
+	// Not really necessary if disabling compression
+	opts.BlockCacheSize = 0
+	// compressions reduces storage usage but increases memory consumption, specially during compaction
+	opts.Compression = options.None
```
mtsitrin, line 230: no compression at all?
maybe we need to separate between sequencer/full node/archive

store/badger.go
srene: channel added as an exit case

store/badger.go
srene: the problem with dymint store is not the storage space but the memory consumption when compacting. its true its not optimal (the size it can be doubled when using no compression) but taking into account that in an unpruned node the application db space can be more than 10x the dymint storage space, it does not seem to be the issue. also this is a short-term solution to avoid oom issues in nodes. long-term solution will probably be replace badger by a more optimal store for dymint requirements.

store/badger.go
mtsitrin: I thought the db has exposed ctx.
from the docs:
```
// Only one GC is allowed at a time. If another value log GC is running, or DB
// has been closed, this would return an ErrRejected.
```

maybe it's more elegenat to just check for this type of error

srene: but if we check for ErrRejected, it maybe the previous didnt finish for some reason, for instance, and we kill the loop while it was not necessary...

mtsitrin: fine
Even though I'm not sure how `maybe the previous didnt finish` is possible. does `RunValueLogGC` actually runs in the background?

mtsitrin (approved):

danwt (dismissed):
The closing channel is a bit weird, that's the point of context, which is what was there before

so context would make more sense?

store/badger.go
srene: yes, sorry, this shoudlnt happen, but from the docs not sure if ErrRejected will only happen when closing the db...

omritoptix (approved):

PR #1079, opened 2024-09-18
chore(rollapp): added x/rollapp proto files

### LINKED ISSUE #1078: Do not use copied autogenerated files for x/rollapp
Dymint imported the Hub previously, but now Cosmos versions are different as well as other deps, so direct import doesn't work currently. As a workaround, the autogenerated proto contracts were copied from the Hub to Dymint. Copying autogenerated files is not safe (it can lead to dependencies mismatch), it's impossible to find what was the original proto for these files, and it's extremely inobvious how to add new fields to such proto contracts (apparently find a suitable generator, generate files using it, and then manually copy them?).

This PR moves x/rollapp proto contracts from the Hub to Dymint, so the files are generated directly in the Dymint repo with its own proto generator.

keruch:
# PR Standards

Close #1078 

Dymint imported the Hub previously, but now Cosmos versions are different as well as other deps, so direct import doesn't work currently. As a workaround, the autogenerated proto contracts were copied from the Hub to Dymint. Copying autogenerated files is not safe (it can lead to dependencies mismatch), it's impossible to find what was the original proto for these files, and it's extremely inobvious how to add new fields to such proto contracts (apparently find a suitable generator, generate files using it, and then manually copy them?).

This PR moves x/rollapp proto contracts from the Hub to Dymint, so the files are generated directly in the Dymint repo with its own proto generator.

---

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


mtsitrin:
@keruch awesome!

can u do the same for the `x/sequencer` and get rid of the `third_party/dymension/sequencer` copied files?

mtsitrin (approved):

PR #1093, opened 2024-09-26
feat: consensus messages

faultytolly:
# PR Standards

ADR link https://www.notion.so/dymension/ADR-x-Sequencer-Messages-34f95e2b579e458e8bb4252da19324bd

Closes https://github.com/dymensionxyz/epics/issues/1

---

---

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -301,8 +301,7 @@ replace (
 	github.com/evmos/evmos/v12 => github.com/dymensionxyz/evmos/v12 v12.1.6-dymension-v0.3
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.2-alpha.regen.4
 	github.com/gorilla/rpc => github.com/dymensionxyz/rpc v1.3.1
-	// github.com/tendermint/tendermint => github.com/dymensionxyz/cometbft v0.34.29-0.20240906143736-1e3959c2826e
-	github.com/tendermint/tendermint => ../cometbft
+	github.com/tendermint/tendermint => github.com/dymensionxyz/cometbft v0.34.29-0.20241008141942-63af9d24107f
 )
 
 replace github.com/osmosis-labs/osmosis/v15 => github.com/dymensionxyz/osmosis/v15 v15.2.0-dymension-v1.1.2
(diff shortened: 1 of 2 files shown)


block/executor.go
```diff
@@ -92,10 +105,26 @@ func (e *Executor) InitChain(genesis *tmtypes.GenesisDoc, valset []*tmtypes.Vali
...
+	lastHeaderHash, nextSeqHash [32]byte,
+	state *types.State,
+	maxBlockDataSizeBytes uint64,
+) *types.Block {
 	maxBlockDataSizeBytes = min(maxBlockDataSizeBytes, uint64(max(minBlockMaxBytes, state.ConsensusParams.Block.MaxBytes)))
 	mempoolTxs := e.mempool.ReapMaxBytesMaxGas(int64(maxBlockDataSizeBytes), state.ConsensusParams.Block.MaxGas)
 
+	var consensusAnyMessages []*proto.Any
+	if e.consensusMessagesStream != nil {
+		consensusMessages, err := e.consensusMessagesStream.GetConsensusMessages()
```
omritoptix, line 120: I think that it would be useful to have a filter by height because I assume the hub most likely would like to sometimes target specific heights (e.g for rollback)

omritoptix:
@faultytolly build fails

types/block.go
```diff
@@ -83,6 +84,7 @@ type Data struct {
 	Txs                    Txs
 	IntermediateStateRoots IntermediateStateRoots
 	Evidence               EvidenceData
+	ConsensusMessages      []*proto.Any
```
omritoptix, line 87: @faultytolly and @testinginprod I think we discussed that we need  a way for full nodes to validate correctness of last consensus msgs they recive and we discussed if we should add a specific ConsensusMessagesHash but we agreed on it being part of the header hash, but it doesn't seen like part of the header and it's not being part of the LastDataHash. 

We can leave it as an open issue for now but just making sure this is a concious decision vs a miss.

omritoptix (commented):
looks good. few minor comments/questions.

types/block.go
omritoptix: adding to this - we do check consensusMsgs Response as its part of the blockResultsHash. so I guess it's fine for now.

omritoptix (approved):

PR #1129, opened 2024-10-10
feat(rpc): new endpoint for block validation status

### LINKED ISSUE #1128: add rpc endpoint to check validation status

srene:
# PR Standards

--

---

Close #1128

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/rpc/client/client.go
+++ b/rpc/client/client.go
@@ -802,11 +802,8 @@ func (c *Client) CheckTx(ctx context.Context, tx tmtypes.Tx) (*ctypes.ResultChec
 }
 
 func (c *Client) BlockValidated(ctx context.Context, height *int64) (*ResultBlockValidated, error) {
+
 	if *height < 0 {
-		return &ResultBlockValidated{Result: -1}, nil
-	}
-	// node is still syncing
-	if c.node.BlockManager.State.Height() < c.node.BlockManager.LastSubmittedHeight.Load() {
 		return &ResultBlockValidated{Result: -1}, nil
 	}
 	// node has not reached the height yet
@@ -816,6 +813,8 @@ func (c *Client) BlockValidated(ctx context.Context, height *int64) (*ResultBloc
 	if uint64(*height) <= c.node.BlockManager.State.GetLastValidatedHeight() {
 		return &ResultBlockValidated{Result: 2}, nil
 	}
+
+	// block is applied, and therefore it is validated at block level but not at state update level
 	return &ResultBlockValidated{Result: 1}, nil
 }
 
--- a/rpc/client/client.go
+++ b/rpc/client/client.go
@@ -36,6 +36,14 @@ const (
 	subscribeTimeout = 5 * time.Second
 )
 
+type ValidationStatus int
+
+const (
+	NotValidated = iota
+	P2PValidated
+	SLValidated
+)
+
 // ErrConsensusStateNotAvailable is returned because Dymint doesn't use Tendermint consensus.
 var ErrConsensusStateNotAvailable = errors.New("consensus state not available in Dymint")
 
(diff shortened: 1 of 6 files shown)


rpc/client/client.go
```diff
@@ -817,6 +821,24 @@ func (c *Client) CheckTx(ctx context.Context, tx tmtypes.Tx) (*ctypes.ResultChec
 	return &ctypes.ResultCheckTx{ResponseCheckTx: *res}, nil
 }
 
+// BlockValidated returns the settlement validation status for a specific height: 0: Not validated, 1: Validated from P2P, 2: Validated from settlement, -1: node is still syncing.
```
omritoptix, line 824: should probably an enum vs hardcoded

rpc/client/client.go
```diff
@@ -53,6 +61,10 @@ type Client struct {
 	genChunks []string
 }
 
+type ResultBlockValidated struct {
+	Result int
```
danwt, line 65: why not use the custom type here?

rpc/json/service.go
```diff
@@ -193,6 +194,7 @@ func (s *service) NetInfo(req *http.Request, args *netInfoArgs) (*ctypes.ResultN
 }
 
 func (s *service) BlockchainInfo(req *http.Request, args *blockchainInfoArgs) (*ctypes.ResultBlockchainInfo, error) {
+	fmt.Println(args)
```
zale144, line 197: remove

srene: done

srene:
> I'm not exactly sure in what context this is used, but might it be desirable for the caller to also pass the header hash to check that the block the caller refers to is the same one the node is referring to?
> 
> To stop the caller and the callee confusing two different blocks with the same height
> 
> Basically the caller wants to know he has the same block the verifier has? or not?

that is a good point but i dont think that the software that requires this rpc has the block itself with the hash, just the height.

rpc/client/client.go
```diff
@@ -817,6 +829,24 @@ func (c *Client) CheckTx(ctx context.Context, tx tmtypes.Tx) (*ctypes.ResultChec
 	return &ctypes.ResultCheckTx{ResponseCheckTx: *res}, nil
 }
 
+// BlockValidated returns the settlement validation status for a specific height: 0: Not validated, 1: Validated from P2P, 2: Validated from settlement, -1: node is still syncing.
+func (c *Client) BlockValidated(ctx context.Context, height *int64) (*ResultBlockValidated, error) {
```
danwt, line 833: why is height a pointer?

danwt (changes requested):

rpc/client/client.go
srene: i think its the common way used in all endpoints. it can be useful  to check if height its set in the input params.

rpc/client/client.go
```diff
@@ -817,6 +829,25 @@ func (c *Client) CheckTx(ctx context.Context, tx tmtypes.Tx) (*ctypes.ResultChec
...
+	// invalid height
+	if *height < 0 || height == nil {
```
zale144, line 835: this can panic. for short circuit evaluation order is important, the OR is evaluated from left to right

```suggestion
	if height == nil || *height < 0 {
```

rpc/client/client_test.go
```diff
@@ -295,6 +295,83 @@ func TestGetBlock(t *testing.T) {
...
+
+	tests := []struct {
+		name            string
+		validatedHeight uint64
+		nodeHeight      uint64
+		submittedHeight uint64
+		queryHeight     int64
+		result          client.ValidationStatus
+	}{
+		{
```
zale144, line 318: would it make sense to add a test case with `queryHeight: nil` ?

settlement/settlement.go
```diff
@@ -32,6 +34,15 @@ type BatchMetaData struct {
 	DA *da.DASubmitMetaData
 }
 
+type BlockDescriptor struct {
```
zale144, line 37: is this a duplicate?

zale144 (changes requested):

rpc/client/client.go
srene: fixed

rpc/client/client_test.go
srene: test added

settlement/settlement.go
srene: yes, removed

rpc/client/client.go
```diff
@@ -36,6 +36,14 @@ const (
 	subscribeTimeout = 5 * time.Second
 )
 
+type ValidationStatus int
+
+const (
+	BlockNotValidated = iota
+	BlockValidated
+	BlockValidatedWithSL
+)
```
omritoptix, line 45: imo better to stick to p2p as `BlockValidated` could be amibigous. 

```suggestion
type BlockValidationStatus int

const (
	NotValidated = iota
	P2PValidated
	SLValidated
)
```

omritoptix (approved):

PR #1130, opened 2024-10-10
feat(manager): state update mishbehavior detection

### LINKED ISSUE #1106: Detection: State Update Misbehavior - Impl
Should be based on the state update misbehavior ADR

srene:
# PR Standards

[ADR x: State update  misbehavior detection](https://www.notion.so/dymension/ADR-x-State-update-misbehavior-detection-112a4a51f86a80d49d4fdcfb2535b15c?pvs=4)

--

---

Close #1106 

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/da/celestia/celestia.go
+++ b/da/celestia/celestia.go
@@ -17,7 +17,6 @@ import (
 	"github.com/tendermint/tendermint/libs/pubsub"
 
 	openrpc "github.com/celestiaorg/celestia-openrpc"
-	"github.com/celestiaorg/celestia-openrpc/types/share"
 
 	"github.com/dymensionxyz/dymint/da"
 	celtypes "github.com/dymensionxyz/dymint/da/celestia/types"
@@ -349,49 +348,6 @@ func (c *DataAvailabilityLayerClient) retrieveBatches(daMetaData *da.DASubmitMet
 		BaseResult: da.BaseResult{
 			Code:    da.StatusSuccess,
 			Message: "Batch retrieval successful",
-		},
-		Batches: batches,
-	}
-}
-
-// RetrieveBatches gets a batch of blocks from DA layer.
-func (c *DataAvailabilityLayerClient) retrieveBatchesNoCommitment(dataLayerHeight uint64) da.ResultRetrieveBatch {
-	ctx, cancel := context.WithTimeout(c.ctx, c.config.Timeout)
-	defer cancel()
-	blobs, err := c.rpc.GetAll(ctx, dataLayerHeight, []share.Namespace{c.config.NamespaceID.Bytes()})
-	if err != nil {
-		return da.ResultRetrieveBatch{
-			BaseResult: da.BaseResult{
-				Code:    da.StatusError,
-				Message: err.Error(),
-			},
-		}
-	}
-
-	var batches []*types.Batch
-	for i, blob := range blobs {
-		var batch pb.Batch
-		err = proto.Unmarshal(blob.Data, &batch)
-		if err != nil {
-			c.logger.Error("Unmarshal block.", "daHeight", dataLayerHeight, "position", i, "error", err)
-			continue
-		}
-		parsedBatch := new(types.Batch)
-		err := parsedBatch.FromProto(&batch)
-		if err != nil {
-			return da.ResultRetrieveBatch{
-				BaseResult: da.BaseResult{
-					Code:    da.StatusError,
-					Message: err.Error(),
-				},
-			}
-		}
-		batches = append(batches, parsedBatch)
-	}
-
-	return da.ResultRetrieveBatch{
-		BaseResult: da.BaseResult{
-			Code: da.StatusSuccess,
 		},
 		Batches: batches,
 	}


block/manager.go
```diff
@@ -226,6 +234,12 @@ func (m *Manager) Start(ctx context.Context) error {
 	/* ----------------------------- full node mode ----------------------------- */
 	if !isProposer {
 
+		// update latest finalized height
+		err = m.updateLastFinalizedHeightFromSettlement()
```
omritoptix, line 238: why do we need to update it explicitly if it's gonna be updated on the next event anyway?

block/validate.go
```diff
@@ -60,3 +68,19 @@ func (m *Manager) SettlementValidateLoop(ctx context.Context) error {
...
+	drsHistory, err := m.Store.LoadDRSVersionHistory()
+	if errors.Is(err, gerrc.ErrNotFound) {
+		m.logger.Info("failed to find drs history in the store, creating new")
+		m.DRSVersionHistory = &types.DRSVersionHistory{}
+		return nil
+	} else if err != nil {
+		return fmt.Errorf("get drs version history: %w", err)
+	}
+
+	m.DRSVersionHistory = drsHistory
```
omritoptix, line 83: as to my other comment on the applyBlock - not sure the optimization of saving it in memory and than only saving the uniqe changes in store (which I'm not sure is actually the case) is necessary for optimization. 
I'd simply just save the DRS per height and also get rid of the new manager memeber (i.e `DRSVersionHistory`)

block/manager.go
```diff
@@ -161,6 +164,11 @@ func NewManager(
 		return nil, fmt.Errorf("get initial state: %w", err)
 	}
 
+	err = m.LoadDrsHistoryFromStore()
```
omritoptix, line 167: as to my other comment - don't think maintining this in memory is necessary

block/block.go
```diff
@@ -122,7 +122,15 @@ func (m *Manager) applyBlock(block *types.Block, commit *types.Commit, blockMeta
 				m.logger.Error("pruning channel full. skipping pruning", "retainHeight", retainHeight)
 			}
 		}
-
+		if responses.EndBlock.RollappParamUpdates != nil {
+			newDRS := m.DRSVersionHistory.AddDRSVersion(block.Header.Height, responses.EndBlock.RollappParamUpdates.Version)
+			if newDRS {
+				_, err = m.Store.SaveDRSVersionHistory(m.DRSVersionHistory, nil)
```
omritoptix, line 128: I think we should simply save DRS per height in the store (so the key is per height, similiar to how we save the proposer changes. And prune it when necessary. it will simplify all the DRS management and validation imo)

omritoptix (changes requested):
I think the DRS in-memory management makes the code cumbersome and unnecessary complicated vs just saving DRS per height (like we do with `block`, `commit`, `proposer`) etc and just get it where necessary. 

we should strive to keep the code simple vs micro optimizations.

block/manager.go
srene: because otherwise on starting a new node (no previous finalized events seen), or when nodes been down for a while, it may start validating already finalized heights when not necssary

block/validate.go
```diff
@@ -60,3 +68,19 @@ func (m *Manager) SettlementValidateLoop(ctx context.Context) error {
...
+
+	return nil
+}
```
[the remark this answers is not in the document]
srene: this has been removed

block/block.go
```diff
@@ -122,7 +122,15 @@ func (m *Manager) applyBlock(block *types.Block, commit *types.Commit, blockMeta
 				m.logger.Error("pruning channel full. skipping pruning", "retainHeight", retainHeight)
 			}
 		}
-
+		if responses.EndBlock.RollappParamUpdates != nil {
+			newDRS := m.DRSVersionHistory.AddDRSVersion(block.Header.Height, responses.EndBlock.RollappParamUpdates.Version)
```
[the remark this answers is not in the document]
srene: drs history replaced by storing drs per height.

block/validate.go
srene: removed

block/manager.go
srene: removed

block/block.go
srene: done

srene: done

block/manager.go
srene: drs history replaced by drs per height in store

block/submit.go
```diff
@@ -201,15 +201,21 @@ func (m *Manager) CreateBatch(maxBatchSize uint64, startHeight uint64, endHeight
 			return nil, fmt.Errorf("load commit: h: %d: %w", h, err)
 		}
 
+		drsVersion, err := m.Store.LoadDRSVersion(block.Header.Height)
```
omritoptix, line 204: I think this is not backward compatible.. we should open an issue on making it backward compatible (don't have to fix it now)

da/celestia/celestia.go
```diff
@@ -328,7 +321,14 @@ func (c *DataAvailabilityLayerClient) retrieveBatches(daMetaData *da.DASubmitMet
...
 	if err != nil {
-		c.logger.Error("Unmarshal block.", "daHeight", daMetaData.Height, "error", err)
+		c.logger.Error("Unmarshal blob.", "daHeight", daMetaData.Height, "error", err)
+		return da.ResultRetrieveBatch{
+			BaseResult: da.BaseResult{
+				Code:    da.StatusError,
+				Message: err.Error(),
+				Error:   da.ErrBlobNotParsed,
+			},
+		}
```
omritoptix, line 331: I now see that we write to celestia a batch which has:
1. LastBatch
2. DRSVersion 

but why is that info needs to be on celestia? 

also not sure it's backward compatible with unmrashaling old batch structures.

omritoptix (approved):
Approved. 2 issues im not sure about but we can address it in seperate issues.

block/submit.go
srene: https://github.com/dymensionxyz/dymint/issues/1170

srene:
da batch is written to celestia without including drs version and last batch, since it is not included when mashall / unmmarshal, but its used when submitting to sl to generate the block descriptors. added an issue to improve it https://github.com/dymensionxyz/dymint/issues/1171

PR #1163, opened 2024-10-23
feat: proposer and sequencer set decoupling

keruch:
# PR Standards

## Opening a pull request should be able to meet the following requirements

ADR link: https://www.notion.so/dymension/Sequencer-Rotation-Internal-audit-124a4a51f86a8066abf5ec64463cf391#125a4a51f86a80c9833cf3ee3a7b8391

ADR checklist:
- [x] Track the non-active sequencer set (in the state but not in the store)
- [X] Sequencer set is moved from the state to the manager
- [X] Sequencer set is deleted from the store
- [x] Sequencer set contains all sequencers the rollapp ever had (bonded + unbonded)
- [x] State keeps the proposer separately
- [x] Save proposer by height in the store / Get proposer by height from the store
- [x] Proposer might be nullable (in both state and store)
- [x] Only return the proposer as a valset in RPC queries -> Dan did it lately  https://github.com/dymensionxyz/dymint/pull/1159
- [x] State Is not used in Execute Block
- [ ] Query lazily the entire set before rotation -> next PR
- [ ] Bonded Sequencer event is Redundant -> next PR

Close 
#1149
https://github.com/dymensionxyz/dymint/issues/1152

For Author:

- [x]  Targeted PR against correct branch
- [x]  included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [x]  Linked to Github issue with discussion and accepted design
- [x]  Targets only one github issue
- [x]  Wrote unit and integration tests
- [x]  All CI checks have passed
- [x]  Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code)

---

For Reviewer:

- [ ]  confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ]  Reviewers assigned
- [ ]  confirmed all author checklist items have been addressed

---

After reviewer approval:

- [ ]  In case targets main branch, PR should be squashed and merged.
- [ ]  In case PR targets a release branch, PR should be rebased.

PR #1167, opened 2024-10-25
feat(validation): sequencer misbehavior detection 

faultytolly:
# PR Standards

--

---

Close #XXX

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


cambios.patch
mtsitrin, line 1: what's this one?

block/slvalidator.go
```diff
@@ -170,6 +170,31 @@ func (v *SettlementValidator) ValidateDaBlocks(slBatch *settlement.ResultRetriev
...
 			return err
 		}
+
+		// we compare the sequencer address between SL state info and DA block
+		// if next sequencer is not set, we check if the sequencer hash is equal to the next sequencer hash
+		// because it did not change. If the next sequencer is set, we check if the next sequencer hash is equal on the
+		// last block of the batch
+		isLastBlock := i == len(slBatch.BlockDescriptors)-1
+		if slBatch.NextSequencer != "" && isLastBlock {
+			nextSequencer := v.blockManager.State.Sequencers.GetByAddress(slBatch.NextSequencer)
```
omritoptix, line 180: I blv this api is deprecated now after @keruch change as we don't directly save the `Sequencers` on the state

omritoptix, line 180: what if we don't have the nextSequencer in the store? we should probably fetch all the sequencers again (we can't count on polling to be exact in terms of syncing). we can use https://github.com/dymensionxyz/dymint/pull/1163/files#diff-c5773c2f4ace59a21cd36af8a0c95b46135151ea34f2b4bc09e6cbecc0c3e46aR169

block/slvalidator.go
```diff
@@ -170,6 +170,31 @@ func (v *SettlementValidator) ValidateDaBlocks(slBatch *settlement.ResultRetriev
...
+			var nextSequencerHash [32]byte
+			copy(nextSequencerHash[:], nextSequencer.MustHash())
+
+			if !bytes.Equal(nextSequencer.MustHash(), daBlocks[i].Header.NextSequencersHash[:]) {
+				return types.NewErrInvalidNextSequencersHashFraud(
+					nextSequencerHash,
+					daBlocks[i].Header.NextSequencersHash,
+				)
+			}
```
omritoptix, line 189: This var seems redundant
```go
var nextSequencerHash [32]byte
```
possible suggestion (Wasn't tested and probably identation errors): 
```suggestion
            nextSequencerHash := nextSequencer.MustHash()
            if !bytes.Equal(nextSequencerHash, daBlocks[i].Header.NextSequencersHash[:]) {
                return types.NewErrInvalidNextSequencersHashFraud(
                    *(*[32]byte)(nextSequencerHash),
                    daBlocks[i].Header.NextSequencersHash,
                )
            }
```

settlement/dymension/events.go
```diff
@@ -111,26 +111,35 @@ func convertToNewBatchEvent(rawEventData ctypes.ResultEvent) (*settlement.EventD
...
 		return nil, errors.Join(errs...)
 	}
 	endHeight := uint64(startHeight + numBlocks - 1)
+
+	var nextProposer string
+	if events["state_update.next_proposer"] == nil {
+		nextProposer = ""
+	} else {
+		nextProposer = events["state_update.next_proposer"][0]
+	}
```
omritoptix, line 136: what is this part used for?
if im not mistaken we only use the new batch to get informed about a new batch and than fetch the full batch from the hub for syncing by it's height (as other fields like block descriptors etc are also absent)

omritoptix (changes requested):

coverage.txt
mtsitrin, line 1: mistake?

block/slvalidator.go
```diff
@@ -146,6 +146,11 @@ func (v *SettlementValidator) ValidateDaBlocks(slBatch *settlement.ResultRetriev
 		return types.NewErrStateUpdateNumBlocksNotMatchingFraud(slBatch.EndHeight, numSLBlocks, numSLBlocks)
 	}
 
+	currentProposer := v.blockManager.State.GetProposer()
+	if currentProposer == nil {
```
mtsitrin, line 150: will it allow historical sync of a halted rollapp? (rollapp without active proposer)

omritoptix: as we're getting this from the state, and the state should reflect the rollapp proposer for height (it's not currently but there is an open issue on it), it means that by the time we've got here the chain is most likely halted (i.e has no current proposer).

in that case we have nothing to validate.

omritoptix (approved):

PR #1172, opened 2024-10-28
fix(settlement): use block descriptors as source of truth for blocks to apply 

### LINKED ISSUE #1169: Make sure block descriptors is the source of truth when applying and validating
Rollback may change the block descriptors so we can't blindly count that it matches da blocks.

srene:
# PR Standards

--

---

Close #1169 

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/block/retriever.go
+++ b/block/retriever.go
@@ -29,9 +29,9 @@ func (m *Manager) ApplyBatchFromSL(slBatch *settlement.Batch) error {
 			if block.Header.Height != m.State.NextHeight() {
 				continue
 			}
-			// We dont apply a block if not included in the block descriptor (adds support for ro)
-			if slBatch.BlockDescriptors[blockIndex].Height != block.Header.Height {
-				continue
+			// We dont apply a block if not included in the block descriptor (adds support for rollback)
+			if blockIndex >= len(slBatch.BlockDescriptors) {
+				break
 			}
 			// We dont validate because validateBlockBeforeApply already checks if the block is already applied, and we don't need to fail there.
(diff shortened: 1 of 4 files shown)


block/retriever.go
```diff
@@ -6,38 +6,45 @@ import (
...
 
 	m.retrieverMu.Lock()
 	defer m.retrieverMu.Unlock()
 
 	var lastAppliedHeight float64
+	blockIndex := 0
 	for _, batch := range batchResp.Batches {
 		for i, block := range batch.Blocks {
+			// We dont apply a block if not included in the block descriptor (adds support for rollback)
+			if blockIndex >= len(slBatch.BlockDescriptors) {
```
omritoptix, line 30: wouldn't it be cleaner to check `block.Header.Height` against the heights height of `BlockDescriptor.Height` so in that case you don't need to add any index etc?

omritoptix (approved):
approved with small suggestion

block/retriever.go
srene: i think the index is necessary anyway, if we want to keep compatiblity with  multiple batches in the same state update

srene: although not supported right now...

omritoptix: another problem I see now that we only increment the `blockIndex` at the end of the loop but we have `continue` in the middle which will make the block index incorrect..

srene: fixed

omritoptix (approved):

PR #1147, opened 2024-10-18
fix(manager/indexer): pruning fixes 

srene:
# PR Standards

## Opening a pull request should be able to meet the following requirements

--

PR naming convention: https://hackmd.io/@nZpxHZ0CT7O5ngTp0TP9mg/HJP_jrm7A

---

Close #1145 

<-- Briefly describe the content of this pull request -->

For Author:

- [ ]  Targeted PR against correct branch
- [ ]  included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ]  Linked to Github issue with discussion and accepted design
- [ ]  Targets only one github issue
- [ ]  Wrote unit and integration tests
- [ ]  All CI checks have passed
- [ ]  Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code)

---

For Reviewer:

- [ ]  confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ]  Reviewers assigned
- [ ]  confirmed all author checklist items have been addressed

---

After reviewer approval:

- [ ]  In case targets main branch, PR should be squashed and merged.
- [ ]  In case PR targets a release branch, PR should be rebased.

PR #1182, opened 2024-10-30
feat(manager): decouple drs version from commit

srene:
# PR Standards

--

---

Close #XXX

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


block/manager.go
```diff
@@ -383,8 +384,12 @@ func (m *Manager) UpdateTargetHeight(h uint64) {
 
 // ValidateConfigWithRollappParams checks the configuration params are consistent with the params in the dymint state (e.g. DA and version)
 func (m *Manager) ValidateConfigWithRollappParams() error {
-	if version.Commit != m.State.RollappParams.Version {
-		return fmt.Errorf("binary version mismatch. rollapp param: %s binary used:%s", m.State.RollappParams.Version, version.Commit)
+	dymintVersion, err := strconv.ParseUint(version.DRSVersion, 10, 64)
```
omritoptix, line 387: I think it's confusing to call it dymint version - cause it's not. it's actually the reason we did this - to not couple it with dymint.
I think `DRSVersion` is fine.

omritoptix (approved):
small var name change

block/manager.go
srene: renamed

block/manager.go
```diff
@@ -383,8 +384,12 @@ func (m *Manager) UpdateTargetHeight(h uint64) {
 
 // ValidateConfigWithRollappParams checks the configuration params are consistent with the params in the dymint state (e.g. DA and version)
 func (m *Manager) ValidateConfigWithRollappParams() error {
-	if version.Commit != m.State.RollappParams.Version {
-		return fmt.Errorf("binary version mismatch. rollapp param: %s binary used:%s", m.State.RollappParams.Version, version.Commit)
+	drsVersion, err := strconv.ParseUint(version.DrsVersion, 10, 32)
+	if err != nil {
+		return fmt.Errorf("unable to parse dymint version")
```
omritoptix, line 389:
```suggestion
		return fmt.Errorf("unable to parse drs version")
```

store/store.go
```diff
@@ -317,20 +317,24 @@ func (s *DefaultStore) LoadValidationHeight() (uint64, error) {
...
-	versionBytes, err := s.db.Get(getDRSVersionKey(height))
+func (s *DefaultStore) LoadDRSVersion(height uint64) (uint32, error) {
+	b, err := s.db.Get(getDRSVersionKey(height))
 	if err != nil {
-		return "", fmt.Errorf("load drs version for height %v: %w", height, err)
+		return 0, err
+	}
+	if err != nil {
+		return uint32(0), fmt.Errorf("load drs version for height %v: %w", height, err)
 	}
```
omritoptix, line 327: duplicate code

```suggestion
	if err != nil {
		return uint64(0), fmt.Errorf("load drs version for height %v: %w", height, err)
	}
```

omritoptix (changes requested):
seems like we remoed some of the sequencer rotation misbehavior.

also not related directly but we should probably change the `PruneBlocks` method name which prunes all the stores to somthing more generic like `Prune`

srene:
> seems like we remoed some of the sequencer rotation misbehavior.
> 
> also not related directly but we should probably change the `PruneBlocks` method name which prunes all the stores to somthing more generic like `Prune`

regarding the first, i dont see it removed in ValidateDaBlocks. regarding the prune function, i renamed in the prunning fix pr, since its more convenient to do it there https://github.com/dymensionxyz/dymint/pull/1147.

omritoptix (approved):

PR #1189, opened 2024-10-30
fix(sync): fixed proposer by height syncing issue on startup and rotation

omritoptix:
# PR Standards

--

---

Close #XXX

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


settlement/dymension/dymension.go
```diff
@@ -312,49 +310,43 @@ func (c *Client) GetLatestFinalizedHeight() (uint64, error) {
 	return res.Height, nil
 }
 
-// GetProposer implements settlement.ClientI.
-func (c *Client) GetProposer() *types.Sequencer {
-	// return cached proposer
-	if !c.proposer.IsEmpty() {
-		return &c.proposer
-	}
+// GetProposerAtHeight return the propoesr at height.
```
omritoptix, line 313:
```suggestion
// GetProposerAtHeight return the proposer at height.
```

block/initchain.go
```diff
@@ -3,13 +3,17 @@ package block
...
+	"fmt"
 
 	tmtypes "github.com/tendermint/tendermint/types"
 )
 
 func (m *Manager) RunInitChain(ctx context.Context) error {
-	// get the proposer's consensus pubkey
-	proposer := m.SLClient.GetProposer()
+	// Get the proposer at the initial height
+	proposer, err := m.SLClient.GetProposerAtHeight(1)
```
srene, line 13: from what i understand, this new function will get the proposer from the state info. in case there is no state info i guess it will return not found, and therefore it will not be possible to run initchain for the sequencer, or other nodes when there  is no state update yet. i guess we need to fallback to default getproposer() when there is no state update yet.

block/manager.go
```diff
@@ -191,14 +200,27 @@ func (m *Manager) Start(ctx context.Context) error {
...
-	// Check if the chain is halted
-	err := m.isChainHalted()
-	if err != nil {
-		return err
+	// Check if a proposer on the rollapp is set. In case no proposer is set on the Rollapp, fallback to the hub proposer (If such exists).
+	// No proposer on the rollapp means that at some point there was no available proposer.
+	// In case there is also no proposer on the hub to our current height, it means that the chain is halted.
+	if m.State.GetProposer() == nil {
+		m.logger.Info("No proposer on the rollapp, fallback to the hub proposer, if available")
+		SLProposer, err := m.SLClient.GetProposerAtHeight(int64(m.State.Height()))
```
srene, line 209: same here, that could fallback to slClient.GetProposer() when not found at specific height (because state info is not submitted yet)

srene: otherwise nodes that are higher of the last state update submitted could not start

settlement/dymension/dymension.go
```diff
@@ -638,3 +633,25 @@ func (c *Client) pollForBatchInclusion(batchEndHeight uint64) (bool, error) {
 
 	return latestBatch.Batch.EndHeight == batchEndHeight, nil
 }
+
+func (c *Client) getLatestProposer() (string, error) {
```
srene, line 637: i'd rather have this func public instead of getting the last one by using height -1, but up to you...

block/initchain.go
omritoptix: I think I'll follow the array logic here where `0` will be the initial proposer and `-1` will be the last. 
we can also have 3 seperate methods (i.e initial, at height, last) but I don't think it's crucuial for now and it's more methods to add to the interface.

srene (approved):

PR #1194, opened 2024-10-31
feat(manager): stop applying blocks after node set unhealthy

srene:
# PR Standards

--

---

Close #XXX

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/block/modes.go
+++ b/block/modes.go
@@ -26,7 +26,7 @@ func (m *Manager) runAsFullNode(ctx context.Context, eg *errgroup.Group) error {
 		return m.SettlementValidateLoop(ctx)
 	})
 
-	m.runFullNodeSubscriptions(ctx)
+	m.subscribeFullNodeEvents(ctx)
 
 	return nil
 }
@@ -70,7 +70,10 @@ func (m *Manager) runAsProposer(ctx context.Context, eg *errgroup.Group) error {
 	return nil
 }
 
-func (m *Manager) runFullNodeSubscriptions(ctx context.Context) {
+func (m *Manager) subscribeFullNodeEvents(ctx context.Context) {
+	if m.AmIProposer() {
+		return
+	}
 	// Subscribe to new (or finalized) state updates events.
 	go uevent.MustSubscribe(ctx, m.Pubsub, "syncLoop", settlement.EventQueryNewSettlementBatchAccepted, m.onNewStateUpdate, m.logger)
 	go uevent.MustSubscribe(ctx, m.Pubsub, "validateLoop", settlement.EventQueryNewSettlementBatchFinalized, m.onNewStateUpdateFinalized, m.logger)
@@ -81,12 +84,12 @@ func (m *Manager) runFullNodeSubscriptions(ctx context.Context) {
 
 }
 
-func (m *Manager) runFullNodeUnSubscriptions(ctx context.Context) {
+func (m *Manager) unsubscribeFullNodeEvents(ctx context.Context) {
 	if m.AmIProposer() {
 		return
 	}
 
-	// Subscribe to new (or finalized) state updates events.
+	// unsubscribe for specific event (clientId)
 	unsubscribe := func(clientId string) {
 		err := m.Pubsub.UnsubscribeAll(ctx, clientId)
 		if err != nil {
--- a/block/modes.go
+++ b/block/modes.go
@@ -71,9 +71,7 @@ func (m *Manager) runAsProposer(ctx context.Context, eg *errgroup.Group) error {
 }
 
 func (m *Manager) subscribeFullNodeEvents(ctx context.Context) {
-	if m.RunMode == RunModeProposer {
-		return
-	}
+
 	// Subscribe to new (or finalized) state updates events.
 	go uevent.MustSubscribe(ctx, m.Pubsub, "syncLoop", settlement.EventQueryNewSettlementBatchAccepted, m.onNewStateUpdate, m.logger)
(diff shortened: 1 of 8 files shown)


mtsitrin (dismissed):

block/modes.go
```diff
@@ -75,3 +69,34 @@ func (m *Manager) runAsProposer(ctx context.Context, eg *errgroup.Group) error {
 
 	return nil
 }
+
+func (m *Manager) subscribeFullNodeEvents(ctx context.Context) {
+	if m.RunMode == RunModeProposer {
```
omritoptix, line 74: probably better to remove to call site

block/modes.go
```diff
@@ -75,3 +69,34 @@ func (m *Manager) runAsProposer(ctx context.Context, eg *errgroup.Group) error {
...
+	go uevent.MustSubscribe(ctx, m.Pubsub, "syncLoop", settlement.EventQueryNewSettlementBatchAccepted, m.onNewStateUpdate, m.logger)
+	go uevent.MustSubscribe(ctx, m.Pubsub, "validateLoop", settlement.EventQueryNewSettlementBatchFinalized, m.onNewStateUpdateFinalized, m.logger)
+
+	// Subscribe to P2P received blocks events (used for P2P syncing).
+	go uevent.MustSubscribe(ctx, m.Pubsub, "applyGossipedBlocksLoop", p2p.EventQueryNewGossipedBlock, m.OnReceivedBlock, m.logger)
+	go uevent.MustSubscribe(ctx, m.Pubsub, "applyBlockSyncBlocksLoop", p2p.EventQueryNewBlockSyncBlock, m.OnReceivedBlock, m.logger)
+}
+
+func (m *Manager) unsubscribeFullNodeEvents(ctx context.Context) {
+	if m.RunMode == RunModeProposer {
```
omritoptix, line 87: same. I'd move to callsite

block/modes.go
```diff
@@ -75,3 +69,34 @@ func (m *Manager) runAsProposer(ctx context.Context, eg *errgroup.Group) error {
...
+	}
+
+	// unsubscribe for specific event (clientId)
+	unsubscribe := func(clientId string) {
+		err := m.Pubsub.UnsubscribeAll(ctx, clientId)
+		if err != nil {
+			m.logger.Error("Unsubscribe", "clientId", clientId, "error", err)
+		}
+	}
+	unsubscribe("syncLoop")
```
omritoptix, line 98: hard-conding the client id can easily be error prone.. I'd suggest change to const.

block/sync.go
```diff
@@ -71,18 +68,16 @@ func (m *Manager) SettlementSyncLoop(ctx context.Context) error {
 
 				settlementBatch, err := m.SLClient.GetBatchAtHeight(m.State.NextHeight())
 				if err != nil {
-					return fmt.Errorf("retrieve batch: %w", err)
+					m.logger.Error("retrieve SL batch", "err", err)
+					break
```
omritoptix, line 72: why do we break vs returning error?

block/sync.go
```diff
@@ -71,18 +68,16 @@ func (m *Manager) SettlementSyncLoop(ctx context.Context) error {
...
 				}
 				m.logger.Info("Retrieved state update from SL.", "state_index", settlementBatch.StateIndex)
 
 				err = m.ApplyBatchFromSL(settlementBatch.Batch)
 				if err != nil {
+					m.freezeNode(context.Background(), err)
```
omritoptix, line 78: why do we freeze node vs returning error and handling on the call site?

block/sync.go
```diff
@@ -92,16 +87,18 @@ func (m *Manager) SettlementSyncLoop(ctx context.Context) error {
 
 				err = m.attemptApplyCachedBlocks()
 				if err != nil {
-					uevent.MustPublish(context.TODO(), m.Pubsub, &events.DataHealthStatus{Error: err}, events.HealthStatusList)
+					m.freezeNode(context.Background(), err)
```
omritoptix, line 90: same here. imo freeze should be called from callsite if possible.

danwt (commented):
Obviously not a blocker, but I think the concept is a bit backwards

A node should be unhealthy if it stops working for whatever reason

So you should turn off the processing, and the unhealthy status should read off of that, not the other way around

block/sync.go
srene: because if it temporary fails the access to SL, do you want to return error and quit the loop? this way it will try again for the next state update but without failing (the rollapp maybe operative and is just an issue with the hub node).

block/sync.go
srene: updated it

block/sync.go
srene: updated it

block/modes.go
srene: updated

block/modes.go
srene: moved to const

block/sync.go
omritoptix: thing is that I think that if we return an error it means the hub node retires exhausted.
in that case it could be due to:
1. bad rpc endpoitn (requires node operator inteference)
2. hub is down

either way I think both should raise node unhleathy and let the operator figure it out

srene:
> Obviously not a blocker, but I think the concept is a bit backwards
> 
> A node should be unhealthy if it stops working for whatever reason
> 
> So you should turn off the processing, and the unhealthy status should read off of that, not the other way around

I created an issue to improve the unhealthy status management this https://github.com/dymensionxyz/dymint/issues/1208

block/sync.go
srene: changed to error

omritoptix (approved):

PR #1205, opened 2024-11-07
fix(hardfork): integration fixes

mtsitrin:
# PR Standards

--

---

Close #XXX

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


jgimeno (approved):

block/produce.go
```diff
@@ -41,7 +41,7 @@ func (m *Manager) ProduceBlockLoop(ctx context.Context, bytesProducedC chan int)
 			return nil
 		case <-ticker.C:
 			// Only produce if I'm the current rollapp proposer.
-			if !m.AmIProposerOnRollapp() {
+			if m.RunMode != RunModeProposer {
```
srene, line 44: why is this modified? is this runMode not always be the same after start? i mean if the produceblockloop is started it already means is the proposer... and the runmode is not updated after start i think.

mtsitrin: so no reason to check against local keys each time again

omritoptix: @mtsitrin this is for when you rotate. 

during rotation you start as the proposer, but you shouldn't produce blocks after you produced your last block yet you should still send your last batch. 

the code you introduced will make the proposer continue producing blocks while he's submitting his last batch in parallel to the new proposer.

PR #1212, opened 2024-11-09
feat(hardfork): improvements on the hardfork pr

srene:
# PR Standards

--

---

Close #XXX

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


proto/types/dymint/state.proto
```diff
@@ -48,6 +48,8 @@ message State {
 
   // Proposer is a sequencer that acts as a proposer. Can be nil if no proposer is set.
   Sequencer proposer = 20;
+
+  int64 version_start_height = 21;
```
mtsitrin, line 52: vesrsion is rollapp revision? or DRS?

types/pb/dymensionxyz/dymension/rollapp/query.pb.go
mtsitrin, line 1: where the proto file for those changes?

types/validation.go
```diff
@@ -67,7 +67,12 @@ func (b *Block) ValidateWithState(state *State) error {
 		return NewErrTimeFraud(b, currentTime)
 	}
 
-	if b.Header.Version.App != state.Version.Consensus.App ||
+	appVersion := state.Version.Consensus.App
+	if b.Header.Height < state.VersionStartHeight {
+		appVersion--
```
mtsitrin, line 72: if this change to support historic sync, how u know the expected revision is -1 and not more?

block/fork.go
```diff
@@ -231,16 +252,16 @@ func (m *Manager) createNewBlocks(consensusMsgs []proto.Message) error {
...
 // preventing duplicate batch submissions.
 //
 // The function performs the following operations:
 //  1. Checks for an existing batch at the specified height via SLClient
 //  2. If no batch exists, creates and submits a new one
-func (m *Manager) ensureBatchExists(height uint64) error {
+func (m *Manager) handleForkBatchSubmission(height uint64) error {
 	resp, err := m.SLClient.GetBatchAtHeight(height)
-	if err != nil {
+	if err != nil && !errors.Is(err, gerrc.ErrNotFound) {
```
mtsitrin, line 264: how u can handle fork instruction, when there's already state info for this height?

proto/types/dymint/state.proto
srene: rollapp revision. i renamed it

types/pb/dymensionxyz/dymension/rollapp/query.pb.go
srene: sorry i didnt manage to create it from protos since i was having some dependencies issues when buidling. i copied them from hub. we need to fix it

types/validation.go
srene: this needs to be changes once we add historic snyc. this is a temporary hack to make it work by now assuming there is a single forkl

block/fork.go
srene: in case state info already exists, it means it was already submitted before and there is no need to submit again. i think this will only happen in case the nodes was stopped while submitting this batch

types/pb/dymensionxyz/dymension/rollapp/query.pb.go
srene: i added the proto to fix it

block/fork.go
mtsitrin: how u got into `handleSequencerForkTransition` in this case?
IMO it should have been checked in `m.forkNeeded()`

go.mod
```diff
@@ -291,12 +291,13 @@ require (
...
 	github.com/whyrusleeping/chunker v0.0.0-20181014151217-fe64bd25879f // indirect
 	go.uber.org/mock v0.4.0 // indirect
-	golang.org/x/tools v0.21.1-0.20240508182429-e35e4ccd0d2d // indirect
+	golang.org/x/tools v0.26.0 // indirect
 	pgregory.net/rapid v1.1.0
 )
 
 replace (
 	github.com/centrifuge/go-substrate-rpc-client/v4 => github.com/availproject/go-substrate-rpc-client/v4 v4.0.12-avail-1.4.0-rc1-5e286e3
+	github.com/dymensionxyz/dymension-rdk => ../dymension-rdk
```
mtsitrin, line 300: remove

settlement/grpc/grpc.go
```diff
@@ -51,6 +51,16 @@ type Client struct {
 	refreshTime    int
 }
 
+func (c *Client) GetRollapp() (*types.Rollapp, error) {
+	// TODO implement me
+	panic("implement me")
```
mtsitrin, line 56: needs to be fixed for e2e tests

types/state_info.go
```diff
@@ -0,0 +1,5 @@
+package types
+
+type StateInfo struct {
```
mtsitrin, line 3: not sure where it's used

types/validation.go
mtsitrin: pls add fixme

types/state_info.go
srene: i dont think its used anymore. removing it

settlement/grpc/grpc.go
srene: i dont think e2e tests are using grpc sl, but implemented just in case..

go.mod
srene: done

block/fork.go
srene: it will enter because the instruction file is not deleted till the batch is submitted (so forkneeded() will be true). so when the node starts again it will enter the handleSequencerForkTransition again, see the block already exists but not the batch and will submit the blocks

types/validation.go
srene: done

danwt (changes requested):
- All the frozen stuff should be done with context cancellation, rather than adhoc
- What is the purpose of frozen compared to shutting down the node? I assume it's to avoid looping daemon restarts. Can you document this?
- Please encapsulate block.header.version.app in one place and everywhere call it by revision
- Revision should go on the batch type. There shouldn't be a need to change method signatures to include revision.
- What is the semantic difference between the revision on the state.version, vs the revision on the state of the last block?
- Let's ditch state.Version. The only relevant thing is the block version. Revision should be first class concept.
- Block ID should be <height,revision> pair, but everything is still stored using just height. Maybe fine for now but I think down the road this is going to make things harder to reason about.
- I don't see what is stopping the full node from using a different DRS than the one the proposer actually uses for the new revision?

block/fork.go
```diff
@@ -0,0 +1,266 @@
...
+			return nil, fmt.Errorf("running faulty DRS version %d", drs)
+		}
+	}
+
+	return []proto.Message{
+		&sequencers.MsgUpgradeDRS{
+			Authority:  authtypes.NewModuleAddress("sequencers").String(),
+			DrsVersion: uint64(actualDRS),
+		},
+	}, nil
```
danwt, line 174: It looks like it's always upgrading DRS? We only want to do it if the last is faulty right

block/fork.go
srene: exactly. documented

block/manager.go
```diff
@@ -184,6 +188,24 @@ func NewManager(
...
+		// Upgrade revision on state
+		state := m.State
+		state.Version.Consensus.App = instruction.Revision
+		state.RevisionStartHeight = instruction.RevisionStartHeight
+		if instruction.RevisionStartHeight == m.State.NextHeight() {
+			drsVersion, err := strconv.ParseUint(version.DrsVersion, 10, 32)
+			if err != nil {
+				return nil, fmt.Errorf("unable to parse drs version")
+			}
+			state.RollappParams.DrsVersion = uint32(drsVersion)
```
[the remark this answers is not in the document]
srene: this needs to be changed to enable the rollapp param validation to pass (which is also done when starting the node) when its necessary to start with the new drs version.  conversion is necessary because, afaik, its only possible to set ldflags using strings. i added conversion func in the same version package.

block/produce.go
```diff
@@ -45,6 +45,11 @@ func (m *Manager) ProduceBlockLoop(ctx context.Context, bytesProducedC chan int)
 				continue
 			}
 
+			// finish the block production loop in case the node is frozen
+			if m.isFrozen() {
+				return nil
+			}
+
```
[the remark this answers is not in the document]
srene: done

block/submit.go
```diff
@@ -48,6 +49,7 @@ func SubmitLoopInner(
 	maxBatchTime time.Duration, // max time to allow between batches
 	maxBatchBytes uint64, // max size of serialised batch in bytes
 	createAndSubmitBatch func(maxSizeBytes uint64) (sizeBlocksCommits uint64, err error),
+	frozen func() bool,
 ) error {
```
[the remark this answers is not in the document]
srene: done

types/validation.go
```diff
@@ -67,7 +67,14 @@ func (b *Block) ValidateWithState(state *State) error {
...
+	// https://github.com/dymensionxyz/dymint/issues/1210
+	appVersion := state.Version.Consensus.App
+	if b.Header.Height < state.RevisionStartHeight {
+		appVersion--
+	}
+
+	if b.Header.Version.App != appVersion ||
 		b.Header.Version.Block != state.Version.Consensus.Block {
 		return ErrVersionMismatch
 	}
```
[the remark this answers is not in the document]
srene: done

block/manager.go
```diff
@@ -184,6 +188,24 @@ func NewManager(
...
 
+	if instruction, forkNeeded := m.forkNeeded(); forkNeeded {
+		// Set proposer to nil
+		m.State.SetProposer(nil)
+
+		// Upgrade revision on state
+		state := m.State
+		state.Version.Consensus.App = instruction.Revision
+		state.RevisionStartHeight = instruction.RevisionStartHeight
+		if instruction.RevisionStartHeight == m.State.NextHeight() {
```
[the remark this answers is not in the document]
srene: no

block/fork.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+// prepareDRSUpgradeMessages prepares consensus messages for DRS upgrades.
+// It performs version validation and generates the necessary upgrade messages for the sequencer.
+//
+// The function implements the following logic:
+//   - If no faulty DRS version is provided (faultyDRS is nil), returns no messages
+//   - Validates the current DRS version against the potentially faulty version
+//   - Generates an upgrade message with the current valid DRS version
+func (m *Manager) prepareDRSUpgradeMessages(faultyDRS []uint32) ([]proto.Message, error) {
```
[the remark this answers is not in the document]
srene: it is necessary to create the upgrade consensus msgs. i dont see the issue with the name

types/validation.go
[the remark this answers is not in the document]
srene: updated

block/submit.go
```diff
@@ -234,7 +242,9 @@ func (m *Manager) SubmitBatch(batch *types.Batch) error {
 	}
 	m.logger.Info("Submitted batch to DA.", "start height", batch.StartHeight(), "end height", batch.EndHeight())
 
-	err := m.SLClient.SubmitBatch(batch, m.DAClient.GetClientType(), &resultSubmitToDA)
+	revision := m.State.Version.Consensus.App
+
+	err := m.SLClient.SubmitBatch(batch, m.DAClient.GetClientType(), &resultSubmitToDA, revision)
```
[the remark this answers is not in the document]
srene: moved to batch

block/fork.go
```diff
@@ -0,0 +1,266 @@
...
+
+	// Create first block with consensus messages
+	if _, _, err := m.ProduceApplyGossipBlock(context.Background(), true); err != nil {
+		return fmt.Errorf("producing first block: %v", err)
+	}
+
+	// Create second empty block
+	if _, _, err := m.ProduceApplyGossipBlock(context.Background(), true); err != nil {
+		return fmt.Errorf("producing second block: %v", err)
+	}
```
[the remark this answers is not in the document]
srene: it does not force to be empty but i dont think forcing empty is necessary since the upgrade happens before the txs are processed.

block/fork.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+// prepareDRSUpgradeMessages prepares consensus messages for DRS upgrades.
+// It performs version validation and generates the necessary upgrade messages for the sequencer.
+//
+// The function implements the following logic:
+//   - If no faulty DRS version is provided (faultyDRS is nil), returns no messages
+//   - Validates the current DRS version against the potentially faulty version
+//   - Generates an upgrade message with the current valid DRS version
+func (m *Manager) prepareDRSUpgradeMessages(faultyDRS []uint32) ([]proto.Message, error) {
```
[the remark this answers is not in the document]
srene: the drs upgrade cnsensus message will be sent in all cases and the rdk will decide to plan an upgrade based on current rollapparams status, because this simplifies dymint flow and it is safer.

srene:
> 1. All the frozen stuff should be done with context cancellation, rather than adhoc
> 2. What is the purpose of frozen compared to shutting down the node? I assume it's to avoid looping daemon restarts. Can you document this?
> 3. Please encapsulate block.header.version.app in one place and everywhere call it by revision
> 4. Revision should go on the batch type. There shouldn't be a need to change method signatures to include revision.
> 5. What is the semantic difference between the revision on the state.version, vs the revision on the state of the last block?
> 6. Let's ditch state.Version. The only relevant thing is the block version. Revision should be first class concept.
> 7. Block ID should be <height,revision> pair, but everything is still stored using just height. Maybe fine for now but I think down the road this is going to make things harder to reason about.
> 8. I don't see what is stopping the full node from using a different DRS than the one the proposer actually uses for the new revision?
> 9. Do we need to document all the fields/state that need to be set correctly by rollback assumption? Where is the information about the actual rollback?

i think we addressed everything but there are things i dont think need to be changed now. ill add issues for what is ncessary:
- state.version  stays the same because it is not something added or modified in this pr and change it now may require a dymint refactor that i think should not be part of this pr.
- as we discussed before, nothing prevents the fullnode to start with a wrong version after rollback, but after applying the first block will fail and it will require rolling back again. that it could be maybe improved in the way we validate rollappparams or  but imo is out of scope of this pr.
- consensus msgs for upgrade will be sent in all cases and the rdk will decide what to do based on current rollapp params status. this avoids overloading the state or adding the previous drs in the instruction, which could lead to errors.

block/fork.go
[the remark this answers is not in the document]
danwt: I see thanks

block/fork.go
```diff
@@ -0,0 +1,275 @@
...
+	ticker := time.NewTicker(LoopInterval) // TODO make this configurable
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-ticker.C:
+			if err := m.checkForkUpdate(ctx); err != nil {
+				continue
```
danwt, line 40: dont you want error here?

block/fork.go
```diff
@@ -0,0 +1,275 @@
...
+			}
+		}
```
danwt, line 42: can dry out and put checkForUpdate at top of for loop?

block/fork.go
```diff
@@ -0,0 +1,275 @@
...
+		},
+	}, nil
+}
+
+// Block Creation Rules:
+//   - Two blocks are considered in this process:
+//   - First block: Contains consensus messages for the fork
+//   - Second block: Should be empty (no messages or transactions)
+//   - Total height increase should be 2 blocks from RevisionStartHeight
+func (m *Manager) handleCreationOfForkBlocks(instruction types.Instruction, consensusMsgs []proto.Message) error {
```
danwt, line 167: just call it create fork blocks?

block/fork.go
```diff
@@ -0,0 +1,275 @@
...
+	if heightDiff == 2 {
+		return m.validateExistingBlocks(instruction, 2)
+	}
+
+	// Case 2: First block exists (heightDiff == 1)
+	if heightDiff == 1 {
+		if err := m.validateExistingBlocks(instruction, 1); err != nil {
+			return err
+		}
+		return m.createNewBlocks(consensusMsgs, 1) // Create only the second block
```
danwt, line 181: why does second block have cons message?

block/fork.go
```diff
@@ -0,0 +1,275 @@
...
+	if blocksToCreate == 2 {
+		if _, _, err := m.ProduceApplyGossipBlock(context.Background(), true); err != nil {
+			return fmt.Errorf("producing first block: %v", err)
+		}
+	}
+
+	// Create second empty block for both cases (blocksToCreate == 1 or 2)
+	if _, _, err := m.ProduceApplyGossipBlock(context.Background(), true); err != nil {
+		return fmt.Errorf("producing second block: %v", err)
+	}
```
danwt, line 250: it's allow empty, it should be forced empty

danwt (changes requested):

block/retriever.go
```diff
@@ -36,6 +37,22 @@ func (m *Manager) ApplyBatchFromSL(slBatch *settlement.Batch) error {
 				continue
 			}
 
+			if block.GetRevision() != m.State.GetRevision() {
+
+				rollapp, err := m.SLClient.GetRollapp()
+				if err != nil {
+					return err
+				}
+				if m.shouldStopNode(rollapp, m.State.GetRevision()) {
```
mtsitrin, line 46: not sure why we check it inside the loop.
a batch has single revision
this can be checked at the start

block/retriever.go
```diff
@@ -36,6 +37,22 @@ func (m *Manager) ApplyBatchFromSL(slBatch *settlement.Batch) error {
...
+					err = m.createInstruction(rollapp)
+					if err != nil {
+						return err
+					}
+					m.freezeNode(context.Background(), fmt.Errorf("syncing to fork height. please restart the node. local_block_height: %d rollapp_revision_start_height: %d local_revision: %d rollapp_revision: %d", m.State.Height(), rollapp.RevisionStartHeight, m.State.GetRevision(), rollapp.Revision))
+					return nil
+				}
```
mtsitrin, line 53: same snippet as `m.checkForkUpdate()`

```suggestion
			m.checkForkUpdate()
```

srene: replaced by checkForkUpdate, adding stop message in the func.

block/retriever.go
srene: because revision is not serialized in the batch, therefore it needs to be obtained from the block in retrieval

mtsitrin: so it should be added in `convertStateInfoToResultRetrieveBatch`

mtsitrin (approved):

PR #1215, opened 2024-11-11
fix(block): da unavailability fixes

mtsitrin:
* if `ApplyBatchFromSL` fails, don't go unhealthy (might be caused due to DA unavailability)
* avoid being stuck in an infinite loop in case of missing DA data

### DIFF
--- a/block/sync.go
+++ b/block/sync.go
@@ -2,10 +2,8 @@ package block
 
 import (
 	"context"
-	"errors"
 	"fmt"
 
-	"github.com/dymensionxyz/gerr-cosmos/gerrc"
 	"github.com/tendermint/tendermint/libs/pubsub"
 
 	"github.com/dymensionxyz/dymint/settlement"
@@ -63,9 +61,6 @@ func (m *Manager) SettlementSyncLoop(ctx context.Context) error {
 					m.logger.Info("Synced from local", "store height", currH, "target height", m.LastSettlementHeight.Load())
 					continue
 				}
-				if !errors.Is(err, gerrc.ErrNotFound) {
-					m.logger.Error("Apply local block", "err", err)
-				}
 
 				settlementBatch, err := m.SLClient.GetBatchAtHeight(m.State.NextHeight())
 				if err != nil {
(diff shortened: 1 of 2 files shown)


srene (approved):
not sure why removing the error log when applying from local

block/sync.go
```diff
@@ -75,7 +70,14 @@ func (m *Manager) SettlementSyncLoop(ctx context.Context) error {
 
 				err = m.ApplyBatchFromSL(settlementBatch.Batch)
 				if err != nil {
-					return fmt.Errorf("process next DA batch. err:%w", err)
+					m.logger.Error("Apply batch from SL", "err", err)
```
omritoptix, line 73: the problem I see with logging and breaking here is that as a node operator you'll never know if you have an issue. 

imagine u use a wrong celestia rpc or the rpc is down and you need to change it.
in that case you simply break and the node operator has no idea that he needs to change it. that's why we prefer to emit health issues.

the rpc retry loop should be long enough to indicate there is an rpc problem and not operator should do something about it (the node operator doesn't look at the logs but have alerts to `health` status)

block/sync.go
```diff
@@ -75,7 +70,14 @@ func (m *Manager) SettlementSyncLoop(ctx context.Context) error {
...
+					break
+				}
+
+				// if height havent been updated, we are stuck
+				// this covers the scenario where no applicable blocks were found in the DA
+				if m.State.NextHeight() == currH {
```
omritoptix, line 79: can u elaborate how this can happen? afaiu we got into the loop because we have new settlement height, which means the blocks are in the da. assuming no fraud (as it's handled by blocks unavaliable in the DA), when will this case happen?

block/sync.go
mtsitrin: ok makes sense
I'll revert

srene: i think that  what we need to  do here is only log  and break in case the error is a da error (ErrRetrieval or ErrBlobNotfound), in the other cases return error as usual. This way if da fails the node will not stop and retry in the next state update but fail in case applyblock actually fails and is not da related.

PR #1218, opened 2024-11-11
fix(validation): added da batch validation against state BD desc

mtsitrin:
refactored nextSequencer validation

# PR Standards

--

---

Close #XXX

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/block/retriever.go
+++ b/block/retriever.go
@@ -47,6 +47,12 @@ func (m *Manager) ApplyBatchFromSL(slBatch *settlement.Batch) error {
 			m.blockCache.Delete(block.Header.Height)
 		}
 	}
+
+	// validate the batch applied successfully and we are at the end height
+	if m.State.Height() != slBatch.EndHeight {
+		return fmt.Errorf("state height mismatch: state height: %d: batch end height: %d", m.State.Height(), slBatch.EndHeight)
+	}
+
 	types.LastReceivedDAHeightGauge.Set(lastAppliedHeight)
 
 	return nil
--- a/block/slvalidator.go
+++ b/block/slvalidator.go
@@ -47,7 +47,7 @@ func (v *SettlementValidator) ValidateStateUpdate(batch *settlement.ResultRetrie
 	for height := batch.StartHeight; height <= batch.EndHeight; height++ {
 		source, err := v.blockManager.Store.LoadBlockSource(height)
 		if err != nil {
-			v.logger.Error("load block source", "error", err)
+			// v.logger.Error("load block source", "error", err)
 			continue
 		}
 
@@ -62,11 +62,9 @@ func (v *SettlementValidator) ValidateStateUpdate(batch *settlement.ResultRetrie
 			continue
 		}
 		p2pBlocks[block.Header.Height] = block
-
 	}
 
 	// load all DA blocks from the batch to be validated
-	daBlocks := []*types.Block{}
 	var daBatch da.ResultRetrieveBatch
 	for {
 		daBatch = v.blockManager.Retriever.RetrieveBatches(batch.MetaData.DA)
@@ -84,8 +82,12 @@ func (v *SettlementValidator) ValidateStateUpdate(batch *settlement.ResultRetrie
 		if errors.Is(checkBatchResult.Error, da.ErrBlobNotIncluded) {
 			return types.NewErrStateUpdateBlobNotAvailableFraud(batch.StateIndex, string(batch.MetaData.DA.Client), batch.MetaData.DA.Height, hex.EncodeToString(batch.MetaData.DA.Commitment))
 		}
+
+		// FIXME: we need infinite loop here? not returning error?
+		continue
 	}
 
+	daBlocks := []*types.Block{}
 	for _, batch := range daBatch.Batches {
 		daBlocks = append(daBlocks, batch.Blocks...)
 	}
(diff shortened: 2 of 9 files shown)


block/slvalidator.go
```diff
@@ -142,13 +144,9 @@ func (v *SettlementValidator) ValidateDaBlocks(slBatch *settlement.ResultRetriev
...
 	numSLBlocks := slBatch.NumBlocks
-	if numSLBlocks != numSlBDs {
-		return types.NewErrStateUpdateNumBlocksNotMatchingFraud(slBatch.EndHeight, numSLBlocks, numSLBlocks)
-	}
-
-	currentProposer := v.blockManager.State.GetProposer()
-	if currentProposer == nil {
-		return fmt.Errorf("proposer is not set")
+	numDABlocks := uint64(len(daBlocks))
+	if numSLBlocks != numSlBDs || numDABlocks != numSlBDs {
```
mtsitrin, line 148: added validation of DA blocks with blockdescriptors

block/slvalidator.go
```diff
@@ -172,36 +170,29 @@ func (v *SettlementValidator) ValidateDaBlocks(slBatch *settlement.ResultRetriev
...
-					daBlocks[i].Header.NextSequencersHash,
-				)
-			}
+	// we compare the sequencer address between SL state info and DA block
+	// if next sequencer is not set, we check if the sequencer hash is equal to the next sequencer hash
+	// because it did not change. If the next sequencer is set, we check if the next sequencer hash is equal on the
+	// last block of the batch
+	lastDABlock := daBlocks[len(daBlocks)-1]
+	expectedNextSeqHash := lastDABlock.Header.SequencerHash
+	if slBatch.NextSequencer != slBatch.Sequencer {
```
mtsitrin, line 181: 1. instead of using state proposer, doing `slBatch.NextSequencer != slBatch.Sequencer` to check for a change
2. skipped `UpdateSequencerSetFromSL` as we can assume sync with SL metadata

block/retriever.go
```diff
@@ -47,6 +47,12 @@ func (m *Manager) ApplyBatchFromSL(slBatch *settlement.Batch) error {
 			m.blockCache.Delete(block.Header.Height)
 		}
 	}
+
+	// validate the batch applied successfully and we are at the end height
```
omritoptix, line 51: if the batch isn't applied succesfully wouldn't we get an err from the `applyBlockWithFraudHandling` ?

omritoptix (approved):
looks good though didn't figure out the extra validation of height vs end height on the retriever.

block/retriever.go
mtsitrin: we're looping over `for i, block := range batch.Blocks`, which data on the DA
it can be empty/missing

block/slvalidator.go
```diff
@@ -47,7 +47,7 @@ func (v *SettlementValidator) ValidateStateUpdate(batch *settlement.ResultRetrie
 	for height := batch.StartHeight; height <= batch.EndHeight; height++ {
 		source, err := v.blockManager.Store.LoadBlockSource(height)
 		if err != nil {
-			v.logger.Error("load block source", "error", err)
+			// v.logger.Error("load block source", "error", err)
```
srene, line 50: i dont think you need to remove this log. this will just happen in case no block is stored for the height, which should not happen, i think, because there will be always a block applied for the validated height.

block/slvalidator.go
```diff
@@ -84,8 +82,12 @@ func (v *SettlementValidator) ValidateStateUpdate(batch *settlement.ResultRetrie
 		if errors.Is(checkBatchResult.Error, da.ErrBlobNotIncluded) {
 			return types.NewErrStateUpdateBlobNotAvailableFraud(batch.StateIndex, string(batch.MetaData.DA.Client), batch.MetaData.DA.Height, hex.EncodeToString(batch.MetaData.DA.Commitment))
 		}
+
+		// FIXME: how to handle non-happy case? not returning error?
+		continue
```
srene, line 87: not sure why the continue at the end of the loop

block/slvalidator.go
```diff
@@ -172,36 +170,29 @@ func (v *SettlementValidator) ValidateDaBlocks(slBatch *settlement.ResultRetriev
...
-				return types.NewErrInvalidNextSequencersHashFraud(
-					daBlocks[i].Header.SequencerHash,
-					daBlocks[i].Header.NextSequencersHash,
-				)
-			}
+	// we compare the sequencer address between SL state info and DA block
+	// if next sequencer is not set, we check if the sequencer hash is equal to the next sequencer hash
+	// because it did not change. If the next sequencer is set, we check if the next sequencer hash is equal on the
+	// last block of the batch
+	lastDABlock := daBlocks[len(daBlocks)-1]
```
srene, line 179: the lastDABlock, shouldnt be the last Bd index? because in case of fork the dablock we should check i guess needs to be the one given by the number of bds included in the sl batch not the da batch.

block/slvalidator.go
mtsitrin: I'll add the log back.
But I think it can happen
if u at height H, and the state info is [H-1, H+100],
u'll have 100 logs of it

block/slvalidator.go
mtsitrin: it was continuing before
I've added the `continue` for clarity
as it's not obvious why we do it in case of error (see the fixme above)

block/slvalidator.go
mtsitrin: good catch. will fix

block/slvalidator.go
srene: im not sure this can happen. if you are at  height h,  and there is a new state update for h+100, then the node will sync first to height h+100, and then validate for the whole state info. upon reception of a new state update where endheight is higher than node height, it first calls the sync loop and from the sync loop it triggers validation once blocks are applied. so always a state info is validated the block should exist locally, unless there is a bug.

mtsitrin: yes u right
it  actually handled by the validation added to the `ApplyBatchFromSL`

srene (approved):

block/slvalidator.go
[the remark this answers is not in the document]
mtsitrin: yeah fixed already

block/retriever.go
```diff
@@ -47,6 +47,12 @@ func (m *Manager) ApplyBatchFromSL(slBatch *settlement.Batch) error {
...
+	if m.State.Height() != slBatch.EndHeight {
```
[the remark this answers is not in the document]
mtsitrin: there’s  `m.retrieverMu.Lock()` in the start of the SL sync

Issue #1138, opened 2024-10-16
add celestia and dym account/balance to prometheus metrics


srene:
basically the idea is that we need to query the relevant accounts for balances (from celestia and hub) and set it to gauges.
this is important so operators will know when their balance is about to finished and can top up.

PR #1247, opened 2024-11-20
feat (rollapp): register balances l2

faultytolly:
# PR Standards

--

---

Close https://github.com/dymensionxyz/dymint/issues/1138

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


da/da.go
```diff
@@ -221,6 +227,9 @@ type DataAvailabilityLayerClient interface {
 
 	// Returns the maximum allowed blob size in the DA, used to check the max batch size configured
 	GetMaxBlobSizeBytes() uint32
+
+	// GetSignerBalance returns the balance for a specific address
+	GetSignerBalance() (*Balance, error)
```
keruch, line 232: is there any reason to return pointer to `Balance` here instead of the value?

da/grpc/grpc.go
```diff
@@ -153,3 +153,8 @@ func (d *DataAvailabilityLayerClient) RetrieveBatches(daMetaData *da.DASubmitMet
 		Batches: batches,
 	}
 }
+
+func (d *DataAvailabilityLayerClient) GetSignerBalance() (*da.Balance, error) {
+	//TODO implement me
+	panic("implement me")
```
keruch, line 159: TODO reminder

da/local/local.go
```diff
@@ -181,3 +181,8 @@ func (m *DataAvailabilityLayerClient) updateDAHeight() {
 func (d *DataAvailabilityLayerClient) GetMaxBlobSizeBytes() uint32 {
 	return maxBlobSize
 }
+
+func (m *DataAvailabilityLayerClient) GetSignerBalance() (*da.Balance, error) {
+	//TODO implement me
+	panic("implement me")
```
keruch, line 187: TODO reminder

settlement/grpc/grpc.go
```diff
@@ -395,3 +395,8 @@ func (c *Client) retrieveBatchAtStateIndex(slStateIndex uint64) (*settlement.Res
 	}
 	return &batchResult, nil
 }
+
+func (c *Client) GetSignerBalance() (*types.Balance, error) {
+	//TODO implement me
+	panic("implement me")
```
keruch, line 401: TODO reminder

settlement/local/local.go
```diff
@@ -344,3 +344,8 @@ func keyFromIndex(ix uint64) []byte {
 	binary.BigEndian.PutUint64(b, ix)
 	return b
 }
+
+func (c *Client) GetSignerBalance() (*types.Balance, error) {
+	//TODO implement me
+	panic("implement me")
```
keruch, line 350: TODO reminder

settlement/settlement.go
```diff
@@ -98,4 +98,6 @@ type ClientI interface {
 	GetRollapp() (*types.Rollapp, error)
 	// GetObsoleteDrs returns the list of deprecated DRS.
 	GetObsoleteDrs() ([]uint32, error)
+	// GetSignerBalance returns the balance of the signer.
+	GetSignerBalance() (*types.Balance, error)
```
keruch, line 102: is there any reason to return pointer to `Balance` instead of the value?

keruch (changes requested):
good!

block/balance.go
```diff
@@ -0,0 +1,100 @@
...
+
+	wg.Wait()
+
+	var errs error
+	if errDA != nil {
+		errs = errors.Join(errs, errDA)
+	}
+	if errSL != nil {
+		errs = errors.Join(errs, errSL)
+	}
```
keruch, line 93: you can do `err := errors.Join(errDA, errSL)` instead and then check `err != nil`. 
`errors.Join()` ignores nil errors: https://go.dev/play/p/8bCiqSJ-HBX.

da/da.go
keruch: it doesn't! this is our interface that we do control, i believe you'll just need to remove a pointer from respective method. let's change to a value receiver then to avoid panics and confusion while reading the code.

block/balance.go
```diff
@@ -0,0 +1,46 @@
...
+	ticker := time.NewTicker(CheckBalancesInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-ticker.C:
+			m.logger.Info("Checking balances.")
+			balances, err := m.checkBalances()
```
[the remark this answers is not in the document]
keruch: yea i understand this, my point is different. imagine this:
1. you get the error from SL get balance, but success in DA
2.  `checkBalances` returns `err` and ideally you want to know why SL returned error
3. but a few lines below in `balance.DA != nil` condition you overwrite the `err` variable when you do `ParseFloat`
4. now you've lost the error from `checkBalances`

that's why i suggest renaming it or checking right away after `checkBalances` (eg logging) to avoid this overwrite.

keruch (changes requested):
linter fails, and some comments from the previous review left

block/modes.go
```diff
@@ -87,6 +87,10 @@ func (m *Manager) runAsProposer(ctx context.Context, eg *errgroup.Group) error {
 	// Monitor and handling of the rotation
 	go m.MonitorProposerRotation(ctx)
 
+	uerrors.ErrGroupGoLog(eg, m.logger, func() error {
+		return m.MonitorBalances(ctx)
+	})
```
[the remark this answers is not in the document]
srene: IMO is better to have balances on both. even though full-nodes dont need to spend any balance, i guess is better to have balances monitored in both cases in case there are rotations...

da/da.go
srene: i dont think its a big issue, but i think it can be dealt inside our wrapper func to the celestia openrpc call and directly return the value in GetSignerBalance

da/grpc/grpc.go
srene: i think we need to implement it otherwise it may cause e2e tests crashing when this is called.

da/local/local.go
srene: same here, i think if this is not implemented it will panic when using mock da

da/grpc/grpc.go
srene: just return 0

keruch (approved):
good!

srene (approved):

PR #1256, opened 2024-11-27
feat: validate genesis bridge data against the hub

keruch:
# PR Standards

- update hub x/rollapp proto contracts with the most recent changes
- validate genesis bridge data against the hub after InitChain

Close #XXX

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


go.mod
```diff
@@ -301,7 +303,7 @@ replace (
 	github.com/evmos/evmos/v12 => github.com/dymensionxyz/evmos/v12 v12.1.6-dymension-v0.3
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.2-alpha.regen.4
 	github.com/gorilla/rpc => github.com/dymensionxyz/rpc v1.3.1
-	github.com/tendermint/tendermint => github.com/dymensionxyz/cometbft v0.34.29-0.20241104165035-feade34f8f89
+	github.com/tendermint/tendermint => ../dymension-cometbft
```
zale144, line 306: remove this

zale144 (changes requested):

go.mod
keruch: done

block/initchain.go
```diff
@@ -32,3 +40,21 @@ func (m *Manager) RunInitChain(ctx context.Context) error {
 
 	return nil
 }
+
+// ValidateGenesisBridgeData validates the genesis bridge data from
+// InitChainResponse against the rollapp genesis stored in the hub.
+func (m *Manager) ValidateGenesisBridgeData(dataBytes []byte) error {
+	if len(dataBytes) == 0 {
```
srene, line 47: not 100% sure, but i think we shouldnt allow this, otherwise we may fail to detect genesis bridge issues in some cases. i guess rdk could return valid data even in mock scenarios by setting the genesis with the required info.

settlement/dymension/dymension.go
```diff
@@ -751,3 +751,31 @@ func (c *Client) GetSignerBalance() (types.Balance, error) {
...
+		Data:      data,
+	}
+
+	err := c.RunWithRetry(func() error {
+		var err error
+		res, err = c.cosmosClient.GetRollappClient().ValidateGenesisBridge(c.ctx, req)
+		return err
+	})
+	if err != nil {
+		return fmt.Errorf("rollapp client: validate genesis bridge: %w", err)
```
srene, line 768: in case it is not possible to validate genesis, for instance because hub rpc node is down, we need to add a clear message that the rollapp node needs to clear the data. even if the genesis is valid, but it is not possible to validate, this will return an error and the node will stop, but when starting again, the node will try to run again initchain but with it fail because initchain already happened, so it is necessary to clear the node to be able to run init chain again.

mtsitrin: good point
maybe we should query the hub for expected value, **before** executing `InitChain`

srene: but you need to run initchain to get the response from rdk module. i guess there are some workarounds to solve it (like retry forever till there is a response, and continue to that point in case of restart, etc) but by now i think is enough by being clear in an error message that, if it fails because not possible to query, it is necessary to clear the state (ie. remove data folder)

keruch: > it will try to run again initchain but it will fail because initchain already happened

I tried it locally and did not see any errors. Anyway, I added a message that one needs to reset the node if there's a genesis validation error.

```go
return fmt.Errorf("Cannot validate genesis bridge data: %w. Please call `$EXECUTABLE dymint unsafe-reset-all` before the next launch to reset this node to genesis state.", err)
```

> maybe we should query the hub for expected value, before executing `InitChain`

I also thought about it. My main concern is that we move validation responsibility from the hub to dymint. Not sure if it's a problem though, but actual genesis validation happens in the hub on IBC transfer, and ideally we want to run the identical logic on InitChain. If we move it to dymint, it might not be identical for some reason. Plus we already almost have an RC in the hub, so don't want to disturb it with new PRs haha

block/initchain.go
keruch: done

srene (approved):

PR #1261, opened 2024-11-29
fix(manager): validate blocks revision when submit

srene:
# PR Standards

--

---

Close #XXX

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


block/submit.go
```diff
@@ -173,6 +173,16 @@ func (m *Manager) CreateAndSubmitBatch(maxSizeBytes uint64, lastBatch bool) (*ty
 	if err != nil {
 		return nil, fmt.Errorf("create batch: %w", err)
 	}
+
+	// check all blocks have the same revision
+	for i, block := range b.Blocks {
+		if len(b.Blocks) <= i+1 {
+			break
```
omritoptix, line 180: can't you just range until one block before the last instead? 
also would add a comment that we validate against the last one.

omritoptix (approved):
approved with one small code suggestion

block/submit.go
srene: i actually moved it inside create batch, so we dont need to iterate again though the blocks to check revision number

PR #1283, opened 2024-12-16
max skew time + batch submission time calculation refactor

### LINKED ISSUE #1282: skew time is using genesis creation time when there is no state update submitted.
this leads to an incorrect skew time when no state update is still submitted. it should use first block creation time when there is no state update.

srene:
# PR Standards

This PR solves two issues:

 - submission time when no state update is submitted should take first block as a reference, and no the genesis time (to calculate skew time).
 - batch_submission_time (config) should be based on submissions time, and no based on block times, otherwise it maybe that the submissions are not exactly every batch_submission_time when no block production. 

Close #1282 ,#1287

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/block/submit.go
+++ b/block/submit.go
@@ -335,7 +335,7 @@ func (m *Manager) GetLastBlockTimeInSettlement() time.Time {
 		return time.Time{}
 	}
 	if errors.Is(err, gerrc.ErrNotFound) {
-		firstBlock, err := m.Store.LoadBlock(uint64(m.Genesis.InitialHeight))
+		firstBlock, err := m.Store.LoadBlock(uint64(m.Genesis.InitialHeight)) //nolint:gosec // height is non-negative and falls in int64
 		if errors.Is(err, gerrc.ErrNotFound) {
 			return time.Now()
 		}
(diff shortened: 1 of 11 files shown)


block/submit.go
```diff
@@ -314,11 +311,36 @@ func (m *Manager) UpdateLastSubmittedHeight(event pubsub.Message) {
 	}
 }
 
+func (m *Manager) GetLastBlockInSettlementTime() (time.Time, error) {
+	lastBlockInSettlement, err := m.Store.LoadBlock(m.LastSettlementHeight.Load())
```
omritoptix, line 315: I think the LastSettlementHeight may get pruned no?

```go
pruningHeight = min(m.NextHeightToSubmit(), uint64(retainHeight))
```

if that's the case and we fallback to first block it may cause unnecessary halt.

srene: true. good catch. i keep last submitted block without pruning to enable skew time calculation and keep code simplicity.

danwt:
https://github.com/dymensionxyz/dymint/issues/1282

omritoptix (dismissed):

omritoptix (approved):

PR #1304, opened 2025-01-06
feat(manager): improve unhealthy status management

### LINKED ISSUE #1208: improve unhealthy status management
right now unhealthy status is set by sending an unhealthy event. this is an issue in case we want to automatically recover from unhealthy status and it is required to refactor the logic to allow that.

mtsitrin:
closes #1208 

Main points in this refactor:
* blockManager's loops runs in error group and return error for unrecoverable error
* on error, block manager will enter unhealthy state
* fraud is handled in single place for main loops

TODOs:
* allow retries for errors in `SettlementSync` and `SettlementValidator` for `m.SLClient.GetBatchAtHeight` errors
* fix `TestLocalBlockWithFraud` (I would consider remove `applyLocalBlock` for good)


block/fork.go
```diff
@@ -283,7 +288,7 @@ func (m *Manager) doForkWhenNewRevision() error {
 
 	// this cannot happen. it means the revision number obtained is not the same or the next revision. unable to fork.
 	if expectedRevision.Number != m.State.GetRevision() {
-		panic("Inconsistent expected revision number from Hub. Unable to fork")
+		return fmt.Errorf("inconsistent expected revision number from Hub (%d != %d). Unable to fork", expectedRevision.Number, m.State.GetRevision())
```
danwt, line 291: params are better at the end but nvm

danwt (dismissed):
Nice, seems a big improvement.

Probs we can further simplify later

Issue #1160, opened 2024-10-22
Remove invalid signature 'fraud'


danwt:
https://github.com/dymensionxyz/dymint/pull/1071 raises a fraud when signature is not valid (relative to proposer pub key)

It's not really fraudable/punishable so it shouldn't raise a fraud, it should just be discarded

PR #1312, opened 2025-01-07
fix(fraud): invalid sig is not a fraud

danwt:
# PR Standards

--

---

Close https://github.com/dymensionxyz/dymint/issues/1160

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/types/errors.go
+++ b/types/errors.go
@@ -14,7 +14,7 @@ import (
 )
 
 var (
-	ErrInvalidSignature      = errors.New("invalid signature")
+	ErrInvalidSignature      = gerrc.ErrInvalidArgument.Wrap("signature")
 	ErrNoStateFound          = errors.New("no state found")
 	ErrEmptyBlock            = errors.New("block has no transactions and is not allowed to be empty")
 	ErrInvalidBlockHeight    = errors.New("invalid block height")
--- a/types/errors.go
+++ b/types/errors.go
@@ -255,33 +255,6 @@ func (e ErrInvalidHeaderHashFraud) Unwrap() error {
 	return gerrc.ErrFault
 }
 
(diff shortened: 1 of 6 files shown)


srene (commented):
Im thinking, that in case of p2p block, its true that invalid signature is not a fraud. 
But in case of block applied from da, can it not be considered a fraud in this case (since we know the block is coming from the sequencer)? 
actually in case of p2p block i think this will never happen because the block will be discarded in p2p block validation because invalid signature and it will never by validated before apply.
Therefore, i think its fine in leaving invalid signature as a fraud, since it will only happen when applying blocks from da.

danwt:
@srene I was thinking similar but I think the code is messier as is right now.
How about a check in ValidateDaBlocks which converts an ErrInvalidSignature to a fraud?

srene:
> @srene I was thinking similar but I think the code is messier as is right now. How about a check in ValidateDaBlocks which converts an ErrInvalidSignature to a fraud?

Imo is better to keep fraud detection for invalid signature in da block. if you think is better to have specific signature check in ValidateDaBlocks, instead of leaving the code as it is, im ok with that.

types/errors.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 var (
-	ErrInvalidSignature      = errors.New("invalid signature")
+	ErrInvalidSignature      = gerrc.ErrInvalidArgument.Wrap("signature")
```
mtsitrin, line 17: maybe set wrap `ErrFault`? so it will trigger the fraud handler

danwt:
Gonna close, it, not worth worrying about

Issue #1221, opened 2024-11-11
Pending submissions metrics don't update when block production paused


Northa:
https://github.com/dymensionxyz/dymint/blob/ed72d9b5c6dd28bb5a02ebb7a536b664facb2121/block/submit.go#L80-L85

During some internal tests i noticed that `rollapp_pending_submissions_skew_batches` metric not updating in real time. 

**Steps to reproduce.** 
1. a working rollapp setup with da-client.
2. restart\stop hub rpc.
3. Simulate a load test by sending some txs into rollapp.
4. Monitor the `rollapp_pending_submissions_skew_batches` metric.

possible fix: https://github.com/Northa/dymint/commit/9c6806638f1e67341557e0825e756b4ff2662b15

UPDATE2:
When it comes to `pause block production` due to `too many batches are pending submission`   metrics are not updating as well. 
Affected metrics
```
rollapp_pending_submissions_skew_bytes
rollapp_pending_submissions_skew_batches
rollapp_pending_submissions_skew_blocks
```   

```time="Nov 12 13:45:31.686" level=error msg="Enough bytes to build a batch have been accumulated, but too many batches are pending submission. Pausing block production until a signal is consumed.[]" module=block_manager```

Not sure but probably metrics must be updated there
https://github.com/dymensionxyz/dymint/blob/30a84dfb75902ed54c6f5a89d663f0f54e0b6af8/block/submit.go#L129-L131

PR #1313, opened 2025-01-08
fix(metrics): various fixes in metrics

mtsitrin:
* moved metrics to dedicated package
* fixed `RollappBlockSizeTxsGauge`
*  renamed `LastReceivedDAHeightGauge` to `LastValidatedHeight`
* removed `RollappPendingSubmissionsSkewBatches` (#1221)
* fixed submission gauges not updating (#1221)


danwt (approved):

block/block.go
```diff
@@ -199,6 +196,12 @@ func (m *Manager) applyBlock(block *types.Block, commit *types.Commit, blockMeta
 		panic("I'm the new Proposer now. restarting as a proposer")
 	}
 
+	// update metrics
```
srene, line 199: is it not better to move them before ValidateConfigWithRollappParams()? it can return err on ValidateConfigWithRollappParams or panic for rotation and metrics wont be updated...

mtsitrin: not sure why it needs to be updated on error
the metrics are not for debugging

PR #1331, opened 2025-01-14
refactor(submit): simplified submit loop

mtsitrin:
better role separation between `trigger` and `submitter` routines


srene (commented):
im thinking this approach may have a flaw. i think that, if a node is restarted for instance 5 min before the next batch submission by time (e.g. batch time 1 hour), it will wait another hour before trying to submit by time, given the ticker is configured using maxBatchSubmitTime. i think in the current approach the ticker is activated more often, and then its checked whether is necessary to submit by time using isLastBatchRecent func. this way, even if a node is restarted, it will submit every 1 hour.

PR #1332, opened 2025-01-14
refactor(block): move da init to NewNode instead of NewManager

mtsitrin:
This allows to spin up a lightweight block manager, without active SL client or DA client


srene (commented):
i think the current solution will not work since it loads the da type from genesis, and it may happen is not defined there (migrated rollapps) or it has changed since genesis definition. IMO there are three options to enable lightweight block manager, without active SL client or DA client.
1 - readd da type in config, and instead of instantiating the da from rollapp param, only validate the da created is the same as the rollapp param on blockmanager start(), and panic in case is not. this way da can be passed to the block manager.
2 - instead of creating the da in the NewManager() moving it to start() func. this way a lightweight block manager can be created without da. 
3 - avoid that da init()  creates any connection (e.g. with da light client), just reading config and init logic. this way it will not be a problem when creating lightweight block manager and it will not fail (it will just require valid config params, but could be default params).

srene:
Closing this PR, since there is another PR refactoring DA client initialization that is already allowing to start block manager without initializing da client https://github.com/dymensionxyz/dymint/pull/1368

PR #1396, opened 2025-03-17
fix(validation): set unhealthy instead of freeze when hub rpc fails in validation

srene:
# PR Standards

--

---

Close #XXX

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:

### DIFF
--- a/block/validate.go
+++ b/block/validate.go
@@ -2,6 +2,7 @@ package block
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/dymensionxyz/dymint/settlement"
 	"github.com/tendermint/tendermint/libs/pubsub"
@@ -35,6 +36,7 @@ func (m *Manager) SettlementValidateLoop(ctx context.Context) error {
 				if err != nil {
 					m.logger.Error("Getting batch from SL", "height", currH, "error", err)
 					if healthy {
+						err := fmt.Errorf("Unable to get state from SL. Err:%w", err)
 						m.setUnhealthy(err)
 					}
 					continue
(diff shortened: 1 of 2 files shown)


block/validate.go
```diff
@@ -32,10 +33,16 @@ func (m *Manager) SettlementValidateLoop(ctx context.Context) error {
...
+					m.logger.Error("Getting batch from SL", "height", currH, "error", err)
+					if m.IsHealthy() {
+						err := fmt.Errorf("Unable to get state from SL. Err:%w", err)
+						m.setUnhealthy(err)
+					}
+					continue
+				}
+				if !m.IsHealthy() {
+					m.setHealthy()
 				}
```
omritoptix, line 45: not sure I understand the logic here..
by this logic, If I can't get the batch than I set node to unhealthy, and skip current validation height, only to try to get next height, which if is ok, than I continue as if nothing happend and set node to healthy again?

so my questions are:
1. what about the previous height? 
2. seems like the error is unrecoverable from the calling function (i.e `GetBatchAtHeight`) so why do we continue?
3. doesnt it make sense that the node actually freezes like it was before? we just need to make sure the operator understand (cause for example if he's using bad RPC not sure we want the node to continue syncing and skip validation - but rather change rpc
)

omritoptix (changes requested):
I actually don't think we need to set unhealthy. I think it makes sense to freeze so the operator will check his hub rpc otherwise he continue without validation.

the only issue I maybe see is if it's not clear enough to the operator that he's node is frozen (i.e if he queries the health status - does he see it?)


if so I think this PR is not required.

srene:
i sort of disagree with that. i think the key point here is to know whether rpc issues are usually recoverable or not.  imo is usually the first so it make sense to recover automatically. in both cases operators they need to check status to see what is the issue. right now, if they dont check the status, they  see the node stopped syncing  and they fix it by restarting (without really detecting the issue). IMO its better to make it recoverable since for RPC nodes what is important is availability.

srene:
we agreed to increase rpc retries to get state info so in case node is frozen we can consider is because is not recoverable. closing this pr.

PR #1413, opened 2025-04-25
feat(da): kaspa da client

### LINKED ISSUE #1418: add Kaspa DA client for dymint

srene:
# PR Standards

PR that implements DA client for KASPA 

ADR: https://www.notion.so/dymension/ADR-Kaspa-DA-client-1eaa4a51f86a80ea9a0fcf29ca767c59?pvs=4

--

---

Close #1418 

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


srene:
> I'm curious did you check this out? https://krc20.gitbook.io/krc20/protocols/data-insertion-mechanism

Not really, i just used plain kaspa tx including payload with data.

da/kaspa/kaspa.go
```diff
@@ -0,0 +1,394 @@
...
+	blob, err := c.client.GetBlob(daMetaData.TxHash)
+	if err != nil {
+		return da.ResultRetrieveBatch{
+			BaseResult: da.BaseResult{
+				Code:    da.StatusError,
+				Message: "Blob not found",
+				Error:   err,
+			},
+		}
+	}
```
omritoptix, line 274: seems like on first failure to fetch the blob we return "blob not found" - but what if the rpc is down? is there a reason no retry mechanism within `GetBlob` ?

da/kaspa/kaspa.go
```diff
@@ -0,0 +1,394 @@
...
+				retry.LastErrorOnly(true),
+				retry.Delay(c.batchRetryDelay),
+				retry.DelayType(retry.FixedDelay),
+				retry.Attempts(c.batchRetryAttempts),
+			)
+			if err != nil {
+				err = fmt.Errorf("broadcast data blob: %w", err)
+				c.logger.Error(err.Error())
+				continue
+			}
```
omritoptix, line 217: can't this logic be simplified like the celestia client where we simply retry forever (which is basically what happens here, only here we retry with backoff, and than retry again):

for reference:

```
daMetaData, err := c.submit(data)
			if err != nil {
				c.logger.Error("Submit blob.", "error", err)
				metrics.RollappConsecutiveFailedDASubmission.Inc()
				backoff.Sleep()
				continue
			}
```

da/kaspa/kaspa.go
```diff
@@ -0,0 +1,394 @@
...
+				retry.Delay(c.batchRetryDelay),
+				retry.DelayType(retry.FixedDelay),
+				retry.Attempts(c.batchRetryAttempts),
+			)
+			if err != nil {
+				c.logger.Error("Check batch availability: submitted batch but did not get availability success status.", "error", err)
+				metrics.RollappConsecutiveFailedDASubmission.Inc()
+				backoff.Sleep()
+				continue
+			}
```
omritoptix, line 240: same as my above comment

da/kaspa/kaspa.go
```diff
@@ -0,0 +1,394 @@
...
+			result = c.checkBatchAvailability(daMetaData)
+			if result.Error == da.ErrBlobNotFound {
+				return retry.Unrecoverable(result.Error)
+			}
+			return result.Error
+		},
+		retry.Attempts(c.batchRetryAttempts), //nolint:gosec // RetryAttempts should be always positive
+		retry.DelayType(retry.FixedDelay),
+		retry.Delay(c.batchRetryDelay),
+	)
```
omritoptix, line 345: few questions:
1. I'm not seeing in the c.checkBatchAvailablity when does `da.ErrBolbNotFound` is returned. is it ever? 
2. we need to differentiate between network error and not found error - is it handled? cause from the code I can't see it.

da/walrus/client_test.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 func TestDataAvailabilityLayerClient(t *testing.T) {
-	//t.Skip("Skipping Walrus client tests")
+	t.Skip("Skipping Walrus client tests")
```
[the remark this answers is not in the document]
srene: i think this test was not supposed to run on CI. i changed it to pass tests. i think its been already changed in main

da/kaspa/kaspa.go
srene: iirc even with error it will not fail, it will loop in the block manager when applying from sl. in the fullnode validation, depending on the error returned, it will freeze the node or it will loop (e.g. when rpc error).

da/kaspa/kaspa.go
srene: i think there is a different case here. IMO we want to retry before failing (that implies a new submission with an additional cost) because it may happen there is some delay between submission and network inclusion, or also some momentary network issues.

da/kaspa/kaspa.go
srene: yes you are rigtht, this part was missing in the code. I added a new case in the client to return da.ErrBolbNotFound when the transaction id included is not found by the kaspa api when getting the blob.

da/kaspa/kaspa.go
omritoptix: yea I agree it should be like that - but afaiu, in the code of `GetBlob` - if there is an rpc error, or not - it will never return `BlobNotFound` error - just the message says `blob not found` but the err itself won't be. 

so in that case full node will loop forever if it's not found?

da/kaspa/client/client.go
```diff
@@ -0,0 +1,247 @@
...
+		}
+
+		balance += utxo.UTXOEntry.Amount()
+	}
+	return balance, nil
+}
+
+// retrieveBlobTx gets Tx, that includes blob parts, using  Kaspa REST-API server (https://api.kaspa.org/docs)
+func (c *Client) retrieveBlobTx(txHash string) (*Transaction, error) {
+	if len(txHash) != TxHashLenght {
```
omritoptix, line 199: typo in `TxHashLenght`

da/kaspa/client/client.go
```diff
@@ -0,0 +1,247 @@
...
+			return
+		}
+	}()
+
+	if resp.StatusCode != 200 {
+		var tx FailedTxRetrieve
+		if err := json.NewDecoder(resp.Body).Decode(&tx); err != nil {
+			return nil, fmt.Errorf("Kaspa transaction decode failed: %w", err)
+		}
+		if tx.Result == "Transaction not found" {
```
omritoptix, line 220: is this a result kaspa returns?

da/kaspa/client/client.go
```diff
@@ -0,0 +1,247 @@
...
+			return nil, da.ErrBlobNotFound
+		}
+		return nil, fmt.Errorf("Http response status code not OK: Status: %d", resp.StatusCode)
+	}
+
+	var tx Transaction
+	if err := json.NewDecoder(resp.Body).Decode(&tx); err != nil {
+		return nil, fmt.Errorf("Kaspa transaction decode failed: %w", err)
+	}
```
omritoptix, line 229: this decoding and tx is repeated twice (also above)

da/kaspa/client/client.go
```diff
@@ -0,0 +1,247 @@
...
+		}
+		if tx.Result == "Transaction not found" {
+			return nil, da.ErrBlobNotFound
+		}
+		return nil, fmt.Errorf("Http response status code not OK: Status: %d", resp.StatusCode)
+	}
+
+	var tx Transaction
+	if err := json.NewDecoder(resp.Body).Decode(&tx); err != nil {
+		return nil, fmt.Errorf("Kaspa transaction decode failed: %w", err)
```
omritoptix, line 228: isn't decoding failed should return `ErrBlobNotParsed`? as we're checking this in sl validator.

da/kaspa/kaspa.go
```diff
@@ -0,0 +1,351 @@
...
+
+	batch := &types.Batch{}
+	err = batch.UnmarshalBinary(blob)
+	if err != nil {
+		c.logger.Error("Unmarshaling blob", "error", err)
+		return da.ResultRetrieveBatch{
+			BaseResult: da.BaseResult{
+				Code:    da.StatusError,
+				Message: "Err Unmarshal",
+				Error:   err,
```
omritoptix, line 268: isn't unmarshaling failure should return `ErrBlobNotParsed`? as we're checking this in sl validator.

omritoptix (changes requested):

da/kaspa/client/client.go
srene: this is not decoding the blob, but the kaspa api response, so it should not be treated as  a fraud coming from unparsable blob...

da/kaspa/client/client.go
srene: not sure what you propose to do different here. i decode the response twice, once when it fails and once for valid tx response (the format is different that is why i use two different struct decode).

PR #1461, opened 2025-09-19
feat: TEE support (cherry from danwt/v1.5.0-rc02-tee0)

danwt:
# PR Standards

--

---

Close #XXX

For Author:

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


block/slvalidator.go
```diff
@@ -29,8 +32,9 @@ func NewSettlementValidator(logger types.Logger, blockManager *Manager) *Settlem
 	}
 
 	validator := &SettlementValidator{
-		logger:       logger,
-		blockManager: blockManager,
+		logger:        logger,
+		blockManager:  blockManager,
+		trustedHeight: blockManager.State.LastBlockHeight.Load(),
```
srene, line 37: why do you use here start height as trusted? i think that will work only in the case the node is not p2p syncing,  but it can also lead to errors in case of using a snapshot where start height does not match last validation.

rpc/client/tee/tee.go
```diff
@@ -0,0 +1,92 @@
...
+
+	lastValidatedHeight := validator.GetLastValidatedHeight()
+	if lastValidatedHeight == 0 {
+		return tee.TEEResponse{}, fmt.Errorf("no blocks validated yet")
+	}
+
+	nonce := rollapptypes.TEENonce{
+		RollappId:       node.BlockManager.State.ChainID,
+		CurrHeight:      lastValidatedHeight,
+		FinalizedHeight: validator.GetTrustedHeight(),
```
srene, line 39: i still think heights included in the tee nonce are confusing and error prone

rpc/client/tee/tee.go
```diff
@@ -0,0 +1,92 @@
...
+	"strings"
+
+	"github.com/dymensionxyz/dymint/node"
+	"github.com/dymensionxyz/dymint/tee"
+	rollapptypes "github.com/dymensionxyz/dymint/types/pb/dymensionxyz/dymension/rollapp"
+)
+
+var (
+	socketPath    = "/run/container_launcher/teeserver.sock"
+	tokenEndpoint = "http://localhost/v1/token"
```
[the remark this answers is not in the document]
danwt: but its not config

block/slvalidator.go
srene: even if its one-shot, im assuming the node may start from snapshot and the stored state will tell you whether the start height is validated or not. even with the assumption where there is no p2p involved (in that case for sure the start height is not validated), it can happen the snapshot was stopped at a point where the last applied block was not validated yet (first the state is applied and then its validated iirc). i understand you can force in the tee setup that this will not happen, but 1) i wouldnt call it trustedHeight() if its only the start height and 2) i dont see a reason not to use `blockManager.Store.LoadValidationHeight()` instead of startHeight, if it will work independently of the snapshot / node state, imo is better.

srene:
Thinking about the `trustedHeight()` issue, i think the way its implemented right now will work if you start the TEE instance from 0, or all states are finalized, but can have issues when starting from snapshot in other cases. In case we want to use any snapshot as TEE instance,  I think the TEE instance must always use the last finalized height as `trustedHeight()`. You can get this either from SL query or passing it as a param from manager (there is already query for finalized height when starting the manager). I think its also necessary to set the `lastValidatedHeight` to last finalized height when in TEE mode, otherwise the snapshot may have already validated some heights (and will be used as reference to submit to Hub). But even if submitted from TEE, not all heights will be validated from TEE...

